PR #1, opened 2021-01-28
Verify dependencies before uploading a package

ilevkivskyi:
This is required for security reasons. cc @JukkaL


scripts/build_wheel.py
```diff
@@ -165,6 +169,25 @@ def collect_setup_entries(
...
 
+def verify_dependency(typeshed_dir: str, dependency: str, uploaded: str) -> None:
+    """Verify this is a valid dependency, i.e. a stub package uploaded by us."""
+    known_distributions = set(os.listdir(os.path.join(typeshed_dir, THIRD_PARTY_NAMESPACE)))
+    dependency = get_version.strip_dep_version(dependency)
+    assert dependency.startswith("types-"), "Currently only dependencies on stub packages are supported"
+    assert dependency[len("types-"):] in known_distributions, "Only dependencies on typeshed stubs are allowed"
+    with open(uploaded) as f:
+        uploaded_distributions = set(f.read().splitlines())
+    assert dependency in uploaded_distributions, f"{dependency} looks like a foreign distribution"
```
JukkaL, line 180: Optional: Give a more specific error message if the dependency is different only because of upper/lower case differences?

scripts/upload_changed.py
```diff
@@ -28,13 +29,21 @@ def main(typeshed_dir: str, commit: str, dry_run: bool = False) -> None:
         if dry_run:
             print(f"Would upload: {distribution}, increment {increment}")
             continue
+        for dependency in build_wheel.read_matadata(
```
JukkaL, line 32: Share this code? It seems repeated.

scripts/upload_some.py
```diff
@@ -25,13 +26,21 @@ def main(typeshed_dir: str, pattern: str) -> None:
         # Setting base version to None, so it will be read from current METADATA.toml.
         increment = get_version.main(typeshed_dir, distribution, version=None)
         increment += 1
+        for dependency in build_wheel.read_matadata(
```
JukkaL, line 29: Here's the other copy which looks the same.

scripts/upload_changed.py
```diff
@@ -28,13 +29,21 @@ def main(typeshed_dir: str, commit: str, dry_run: bool = False) -> None:
...
+            os.path.join(typeshed_dir, build_wheel.THIRD_PARTY_NAMESPACE, distribution)
+        ).get("requires", []):
+            if get_version.check_exists(get_version.strip_dep_version(dependency)):
+                # If this dependency is already present, check it was uploaded by us.
+                build_wheel.verify_dependency(typeshed_dir, dependency, uploaded)
```
JukkaL, line 37: Should we always check that the dependency is on PyPI? Otherwise it may be missing, and somebody could upload a bad distribution afterwards?

scripts/upload_changed.py
ilevkivskyi: Yeah, I tried, but because of one line in the middle that is different it doesn't look worth it.

scripts/upload_changed.py
ilevkivskyi: Now that I sort by dependency order it should be OK.

scripts/upload_changed.py
```diff
@@ -17,24 +18,36 @@
...
         subprocess.run(["twine", "upload", os.path.join(temp_dir, "*")], check=True)
+        build_wheel.update_uploaded(uploaded, distribution)
 
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("typeshed_dir", help="Path to typeshed checkout directory")
     parser.add_argument("previous_commit", help="Previous typeshed commit for which we performed upload")
+    parser.add_argument("uploaded", help="Previously uploaded packages to validate dependencies")
```
JukkaL, line 50: This is a path to a file, not the name of packages? Make the help text more explicit.

scripts/upload_changed.py
JukkaL: What is the `check_exists` check for? Can we verify the dependency unconditionally? If a dependency doesn't exist, we should reject the dependency, since somebody could later upload some bad stuff, I think.

JukkaL (commented):
Thanks for the updates! I'm not still sure if the dependency logic quite right. Can you double check it?

scripts/get_version.py
```diff
@@ -34,6 +37,30 @@ def read_base_version(typeshed_dir: str, distribution: str) -> str:
     return data["version"]
 
 
+def strip_dep_version(dependency: str) -> str:
+    """Strip a possible version suffix, e.g. types-six>=0.1.4 -> types-six."""
```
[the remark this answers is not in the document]
ilevkivskyi: There should not be semicolons, `requires` is a list already.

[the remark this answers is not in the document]
ilevkivskyi: Sorry, I didn't notice what is in the second part, I think we may not support this initially.

PR #3, opened 2021-02-01
Various fixes and logging

ilevkivskyi:
In particular:
* Add typeshed commit to description of generated wheel.
* List all packages to upload before uploading them.
* Commit in force upload action to preserve uploaded list changes.

### DIFF
--- a/.github/workflows/force_update.yml
+++ b/.github/workflows/force_update.yml
@@ -35,3 +35,8 @@ jobs:
         run: |
           cd main
           python -m scripts.upload_some ../typeshed ${{ github.event.inputs.distribution }} data/uploaded_packages.txt
+          if [ -z "$(git status --porcelain)" ]; then
+              exit 0;
+          fi
+          git config user.name github-actions && git config user.email github-actions@github.com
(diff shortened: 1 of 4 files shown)


scripts/build_wheel.py
```diff
@@ -54,6 +54,7 @@
 types and metadata should be contributed there.
 
 See https://github.com/python/typeshed/blob/master/README.md for more details.
+This package is generated from typeshed commit `{commit}`.
```
JukkaL, line 57: Nit: Maybe "was generated"?

.github/workflows/force_update.yml
```diff
@@ -35,3 +35,8 @@ jobs:
         run: |
           cd main
           python -m scripts.upload_some ../typeshed ${{ github.event.inputs.distribution }} data/uploaded_packages.txt
+          if [ -z "$(git status --porcelain)" ]; then
```
JukkaL, line 38: When would we need this? Add a comment?

JukkaL (approved):

PR #6, opened 2021-02-02
Fix dependency sorting

ilevkivskyi:


ilevkivskyi:
Unfortunately, I was not able to find a unit test for this, but I am quite sure the new algorithm is strictly better.

PR #9, opened 2021-04-28
Add license information to package metadata

RobbeSneyders:
This PR adds license information to the generated packages as explained in [typeshed issue #5263](https://github.com/python/typeshed/issues/5263).

As mentioned in the issue, typeshed is licensed partially under Apache and partially under MIT. I opted to add the Apache License since it is the most restrictive of both, but IANAL :)


ilevkivskyi (approved):
I guess the tests should now pass on rebase, so I will just merge it.

PR #10, opened 2021-04-30
Fix tests

srittau:
Use typed-ast instead of typing-extensions as "well-known" package,
since typing-extensions are now considered part of the stdlib.

See https://github.com/python/typeshed/pull/5233.


ilevkivskyi (approved):
Thanks!

PR #11, opened 2021-05-12
Description can now be amended per package

srittau:
The extra_description field in METADATA.toml now gets added
to the package's long_description field.

### DIFF
--- a/scripts/build_wheel.py
+++ b/scripts/build_wheel.py
@@ -53,6 +53,8 @@ that uses `{distribution}`. The source for this package can be found at
 https://github.com/python/typeshed/tree/master/stubs/{distribution}. All fixes for
 types and metadata should be contributed there.
 
+{extra_description}
+
 See https://github.com/python/typeshed/blob/master/README.md for more details.
 This package was generated from typeshed commit `{commit}`.
 '''.lstrip()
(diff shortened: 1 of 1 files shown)


scripts/build_wheel.py
```diff
@@ -53,6 +53,8 @@
 https://github.com/python/typeshed/tree/master/stubs/{distribution}. All fixes for
 types and metadata should be contributed there.
 
+{extra_description}
```
ilevkivskyi, line 56: Maybe only add the extra newlines if the extra description is non-empty?

ilevkivskyi (approved):

scripts/build_wheel.py
srittau: I think that multiple empty lines should result in only one new paragraph in markdown, but I'm not certain that's true and whether all markdown renderers do that.

Should I just split the constant and use string concat?

srittau:
@ilevkivskyi I have rewritten the long description generation, also in preparation of #12. This should avoid extra empty lines.

ilevkivskyi (approved):

scripts/build_wheel.py
ilevkivskyi: How you did with having `parts` looks good.

PR #12, opened 2021-05-12
Packages can now be marked as obsolete

srittau:
Using the obsolete_since field in METADATA.toml, a package
can be marked as obsolete, starting with a certain upstream
package version. For example:

obsolete_since = 2.0.0


scripts/build_wheel.py
```diff
@@ -53,6 +53,8 @@
 https://github.com/python/typeshed/tree/master/stubs/{distribution}. All fixes for
 types and metadata should be contributed there.
 
+{obsolete_text}
```
ilevkivskyi, line 56: Again, maybe add extra newlines only when non-empty?

ilevkivskyi (approved):

srittau:
@ilevkivskyi Merged and reworked as well.

ilevkivskyi (approved):

PR #16, opened 2021-06-12
Skip @tests directories

srittau:


srittau:
This should fix uploading. Cc @hauntsaninja

hauntsaninja (approved):
Thanks for fixing! (Also note I don't have write access to this repo, so can't hit merge)

PR #19, opened 2021-06-16
Create an sdist for every wheel that is built

nemacysts:
Some users prefer to build their own wheels, and the fact that the
modular types packages are only uploaded as wheels means that these
users can't use pip with "--no-binary :all:" to guarantee that pip only
downloads sdists in order to then rebuild the corresponding wheel.

### DIFF
--- a/scripts/build_wheel.py
+++ b/scripts/build_wheel.py
@@ -369,6 +369,7 @@ def main(typeshed_dir: str, distribution: str, increment: int) -> str:
     current_dir = os.getcwd()
     os.chdir(tmpdir)
     subprocess.run(["python3", "setup.py", "bdist_wheel", "--universal"])
+    subprocess.run(["python3", "setup.py", "sdist"])
     os.chdir(current_dir)
     return f"{tmpdir}/dist"
 


nemacysts:
This looks like it should resolve https://github.com/python/typeshed/issues/5643 - although I can't test the rest of the workflow, https://github.com/typeshed-internal/stub_uploader/blob/main/scripts/upload_changed.py#L48 looks like it'd also upload the adist that will now be created.

I tested this by:

1.  creating + activating a venv
2. installing the python packages mentioned in the Github Actions setup into that venv
3. cloning typeshed
4. running `python -m scripts.build_wheel typeshed/ mock 1`
5. verifying that the tmpdir contains both a wheel and an sdist: 
    ```ls /tmp/tmpn6e21lm7/dist
    types-mock-0.1.1.tar.gz  types_mock-0.1.1-py2.py3-none-any.whl
    ```

srittau (approved):
Thank you, let's just try it out.

JelleZijlstra:
Should we trigger an update for all types-* packages somehow so that we get sdists for existing packages?

srittau:
Sounds like a good idea.

ilevkivskyi:
You both should have permission to do this https://github.com/typeshed-internal/stub_uploader/actions/workflows/force_update.yml (IIRC the package accepts regex, so you probably want something like `'.*'`).

ilevkivskyi:
Actually, let me do this right now.

ilevkivskyi:
And it is done! (took around 5min to build and upload all)

JelleZijlstra:
Thanks Ivan!

PR #20, opened 2021-06-18
Only use --universal for combined Python 2 and 3 packages

srittau:


srittau:
This is based on [the documentation](https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels) that says:

> Only use this setting if your project does not have any C extesions _and_ supports Python 2 and 3.

JelleZijlstra (approved):

PR #23, opened 2021-08-23
Create and commit changelogs

srittau:
This is a first step for python/typeshed#5633. The current
implementation generates and commits the changelogs, but
doesn't add them to the generated packages.


srittau:
We should test this for a few days and check whether the generated changelogs look ok. Then we can add the generated changelogs to the uploaded packages.

srittau:
For reviewing: It's best to review the two commits on their own.

JelleZijlstra (approved):

PR #25, opened 2021-09-01
Ignored gitignored files when collecting entries

nipunn1313:
I found that while I was testing this script, it would crash if I had a typeshed file open in my editor. This fixes that, and adds some missing tests for the file collection method (which will be handy later as well).

### DIFF
--- a/data/test_typeshed_stubs/bogusfiles/METADATA.toml
+++ b/data/test_typeshed_stubs/bogusfiles/METADATA.toml
@@ -1,0 +1,2 @@
+version = "0.1"
+python2 = true
--- a/data/test_typeshed_stubs/multifilepkg/METADATA.toml
+++ b/data/test_typeshed_stubs/multifilepkg/METADATA.toml
@@ -1,0 +1,2 @@
+version = "0.1"
+requires = []
--- a/data/test_typeshed_stubs/singlefilepkg/METADATA.toml
+++ b/data/test_typeshed_stubs/singlefilepkg/METADATA.toml
@@ -1,0 +1,2 @@
+version = "0.1"
+python2 = true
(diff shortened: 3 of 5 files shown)


nipunn1313:
For example this unit test might be useful for #24

JelleZijlstra:
I feel like the test files are a bit too much, it will make people think they need to update these stubs or otherwise get confused by them. What do you think of using mostly empty example files instead? It doesn't seem like the stub uploader would care what is in the files.

nipunn1313:
sure - I can make the test cases slimmer and contrived. One sec. Didn't think about the subsequent developer confusion.

nipunn1313:
There it is - much simpler diff

data/test_typeshed_stubs/.gitignore
```diff
@@ -0,0 +1 @@
+.*.swp
```
srittau, line 1: We could add this to the main `.gitignore` file.

nipunn1313: lo and behold - it's already there. Just removing this .gitignore file.

nipunn1313:
Have to add some sorting to the unit tests so it runs consistently on different platforms.

nipunn1313:
(oops - sorry, stepped on your feet while you were running the workflow - sorry)

Realized while writing the unit test that I had only covered singlefile pkgs - so I added coverage and unit test for a multifile pkg with gitignore.

JelleZijlstra:
Don't worry about it, happy to keep pushing the button to allow the workflow to run.

nipunn1313:
All complete now!

srittau (approved):

PR #30, opened 2021-09-30
Don't restrict version numbers to two digits

srittau:
python/typeshed#5992, required for python/typeshed#6094


scripts/get_version.py
```diff
@@ -28,6 +28,23 @@
...
+    with requests.Session() as session:
+        session.mount("https://", HTTPAdapter(max_retries=retry_strategy))
+        resp = session.get(url, timeout=TIMEOUT)
+    if not resp.ok:
+        if resp.status_code == 404:
+            # Looks like this is first time this package is ever uploaded.
+            return []
+        raise ValueError("Error while retrieving version")
+    data = resp.json()
+    return data["releases"].keys()
```
Akuli, line 45: This does not return `list[str]`. You can change the annotation to `Iterable` or change this line to `return list(data["releases"].keys())`.

srittau: Good catch. We should start to type check these scripts at some point.

Akuli: They are already type checked, but `resp.json()` returns `Any`.

Issue #31, opened 2021-10-01
Please explicitly declare a licence to this repository


terceiro:
Hi,

I wanted to use some code from this repository for providing the typeshed stub packages in Debian. For that, I need to have specific licensing terms for ir. Anyone else wanting to reuse the code here also needs that, at least to do it legally.

srittau:
@ilevkivskyi @nipunn1313 @hauntsaninja As contributors with non-trivial changes: Do you agree if we use the Apache License 2.0 that typeshed uses for your code?

hauntsaninja:
Sure, whatever works for you works for me! Fwiw, I seem to remember typeshed is licensed under both Apache and MIT

ilevkivskyi:
Apache license should be OK I think. Also cc @JukkaL for visibility.

PR #37, opened 2021-10-05
Add Apache License

srittau:
Closes: #31

See that issue for contributor consent.


srittau:
Added @JukkaL as reviewer as the only contributor left to agree.

JukkaL (approved):
This makes sense.

PR #40, opened 2021-11-30
Link the packages and the script that generated them

ewjoachim:
This script is responsible for quite a few important packages. This PR suggests adding an explicit link between this script and the packages it produces, so that one would be able to study how it works. I got the info by asking [here](https://github.com/python/typeshed/pull/6454#issuecomment-982674966) (thanks @Akuli)

We can also document the link in the `typeshed` README, and we can decide to do either or both. I would say that having a way to go from a project in PyPI to the code that generated it is always a good idea. That being said, it's not mandatory.

Thanks and have a nice day :)

### DIFF
--- a/scripts/build_wheel.py
+++ b/scripts/build_wheel.py
@@ -84,6 +84,9 @@ types and metadata should be contributed there.
 DESCRIPTION_OUTRO_TEMPLATE = """
 See https://github.com/python/typeshed/blob/master/README.md for more details.
 This package was generated from typeshed commit `{commit}`.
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
I'm open to different opinions, but I think having this in the typeshed README is enough, and I'd prefer to keep the generated descriptions short.

(Also, "daily" is wrong.)

ilevkivskyi:
I don't think we need this, sorry. This repo is for internal use, and shouldn't be advertised widely.

PR #42, opened 2022-01-03
Use the new "Typing :: Stubs Only" classifier

ewjoachim:
The new `Typing :: Stubs Only` classifier in [trove-classifiers==2022.1.3](https://pypi.org/project/trove-classifiers/2022.1.3/) is especially well suited for packages that don't provide typed code, but stubs for code that exists elsewhere.

Note that we'll need to wait for [warehouse](https://github.com/pypa/warehouse) to adopt this version of trove-classifiers before making the change, but this should arrive soon, as the PR will be crafted by dependabot quickly :)

### DIFF
--- a/data/empty_package/setup.py
+++ b/data/empty_package/setup.py
@@ -14,6 +14,6 @@ setup(
     install_requires=[],
     packages=[],
     classifiers=[
-        "Typing :: Typed",
+        "Typing :: Stubs Only",
     ],
 )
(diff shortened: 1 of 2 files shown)


srittau (approved):
Thanks!

ewjoachim:
Halfway there :) https://github.com/pypa/warehouse/pull/10551

ewjoachim:
Classifier now appears in https://pypi.org/classifiers/
Which means everything's ready !

srittau:
I'll upload new versions of all stubs, once #43 is fixed.

ewjoachim:
Thanks !

PR #43, opened 2022-01-05
upload_some: Quote distributions so that patterns work

srittau:


ilevkivskyi (approved):
SG, but you may need to also update README that explicitly recommends quotes for regex input.

PR #44, opened 2022-01-28
Add additional project URLs

srittau:


scripts/build_wheel.py
```diff
@@ -52,6 +52,12 @@
       long_description=long_description,
       long_description_content_type="text/markdown",
       url="https://github.com/python/typeshed",
+      project_urls={
+        "GitHub": "https://github.com/python/typeshed",
```
srittau, line 56: While this is a duplicate with the generic `url` field, it adds a GitHub icon on PyPI, which can be helpful.

scripts/build_wheel.py
```diff
@@ -52,6 +52,12 @@
       long_description=long_description,
       long_description_content_type="text/markdown",
       url="https://github.com/python/typeshed",
+      project_urls={{
+          "GitHub": "https://github.com/python/typeshed",
```
srittau, line 56: While this is a duplicate of the generic url field, it adds a GitHub icon on PyPI, which can be helpful.

srittau:
#45 for the type fix.

PR #47, opened 2022-04-29
Remove Python 2 remnants

srittau:
Also add the Python 3 trove classifier to generated packages


hauntsaninja (approved):
Changes look good, but I didn't grep to see if anything was missing

PR #48, opened 2022-06-02
Add a new metadata key "no_longer_updated"

srittau:
This key can be used to mark a package as obsolete, in cases other than
the upstream package getting type annotations.

### DIFF
--- a/scripts/build_wheel.py
+++ b/scripts/build_wheel.py
@@ -71,7 +71,7 @@ setup(name=name,
 ).lstrip()
 
 NO_LONGER_UPDATED_TEMPLATE = """
-*Note:* The `types-{distribution}` is obsolete and unmaintained.
+*Note:* The `types-{distribution}` is unmaintained and won't be updated.
 """.lstrip()
 
 OBSOLETE_SINCE_TEXT_TEMPLATE = """
--- a/scripts/build_wheel.py
+++ b/scripts/build_wheel.py
@@ -71,11 +71,14 @@ setup(name=name,
 ).lstrip()
 
 OBSOLETE_TEXT_TEMPLATE = """
+*Note:* The `types-{distribution}` is obsolete and unmaintained.
+""".lstrip()
+
+OBSOLETE_SINCE_TEXT_TEMPLATE = """
 *Note:* The `{distribution}` package includes type annotations or type stubs
 since version {obsolete_since}. Please uninstall the `types-{distribution}`
 package if you use this or a newer version.
 """.lstrip()
-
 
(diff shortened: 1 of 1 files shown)


srittau:
I don't follow. In both cases the types package is what is obsolete, although `obsolete_since` would more correctly be called "obsolete since upstream version".

JelleZijlstra:
The types package is obsolete in both cases, but the two have pretty different effects on users ("I should stop using this package" vs. "I should upgrade and remove the types package"). Also, for a typeshed contributor, it would be difficult to remember which is which.

srittau:
I don't what to shuffle `obsolete_since` around, since it's/was actively used in typeshed. `package_unmaintained` would again limit what we can use the key for, I would like to keep this fairly general so we don't need to change it when we want to remove a types package for another reason. Maybe `types_removed` or `to_be_removed`?

srittau:
Bumpity bump.

JelleZijlstra:
Would be good to get some other opinions on naming (@AlexWaygood @hauntsaninja @Akuli?).

I'd be fine with `to_be_removed` though, and agree that it's awkward to rename the existing `obsolete_since` key.

AlexWaygood:
The intent is to say that these stubs will no longer be updated, right? So how about `no_longer_updated`?

AlexWaygood:
> Current options:
> 
> * `to_be_removed`
> * `no_longer_updated`
> 
> Both sound fine to me. Let's not forget that this key is only briefly used, though. Basically, add the key, wait for the release, then we can remove the package. So I'd say the name is not super important as it will usually only be set for a day or two.

I'm also happy with both -- I'd vote for `no_longer_updated`, but no strong objections from me to `to_be_removed` 👍

scripts/build_wheel.py
```diff
@@ -70,13 +70,16 @@
 """
 ).lstrip()
 
-OBSOLETE_TEXT_TEMPLATE = """
+NO_LONGER_UPDATED_TEMPLATE = """
+*Note:* The `types-{distribution}` is obsolete and unmaintained.
```
AlexWaygood, line 74: Maybe something like this? But not a blocking concern:

```suggestion
*Note:* The `types-{distribution}` is unmaintained and cannot now be updated.
```

AlexWaygood (approved):

scripts/build_wheel.py
srittau: I tweaked your wording a bit more.

AlexWaygood: LGTM

hauntsaninja:
(this seems great to me!)

Akuli (approved):

PR #49, opened 2022-07-15
Rename scripts to stub_uploader

hauntsaninja:
These started off as standalone scripts, but were made into a package in 1680034


srittau (approved):
Sounds like a good idea to me.

hauntsaninja:
Gentle bump :-)

PR #57, opened 2022-08-21
Change versioning scheme to be more robust

hauntsaninja:
Currently sqlalchemy breaks typeshed CI basically whenever it releases a
new micro version. This would allow us to pin sqlalchemy to a tighter
version in METADATA.toml, while ensuring users who do
`pip install types-sqlalchemy` will still get the latest stub.

Please see the comments in `compute_incremented_version` about what we
attempt to guarantee (and the corresponding asserts in that function).

There's also some associated refactoring so that we can unit test logic
easily and so that we can lean a little more heavily on `packaging` to do
things for us.

### DIFF
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -71,10 +71,7 @@ def compute_incremented_version(
     # all published versions. This ensures that users who don't pin get the most
     # up to date stub. If we ever maintain multiple versions for a stub, this will
     # need revisiting.
-    max_published = max(
-        published_versions,
-        default=Version("0"),
-    )
+    max_published = max(published_versions, default=Version("0"))
 
     # The second thing we try to do (but don't guarantee), is that the incremented
     # version will satisfy the version_spec (defined precisely by the `compatible`
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -60,7 +60,7 @@ def check_exists(distribution: str) -> bool:
     raise ValueError("Error while verifying existence")
 
 
-def extend_to_specificity(ver: list[int], specificity: int) -> None:
+def ensure_specificity(ver: list[int], specificity: int) -> None:
     ver.extend([0] * (specificity - len(ver)))
 
 
@@ -105,7 +105,7 @@ def compute_incremented_version(
         # For example, version_base=1.2, max_published=1.3.0.4, return 1.3.0.5
         increment_specificity = max(increment_specificity, len(max_published.release))
         incremented = list(max_published.release)
-        extend_to_specificity(incremented, increment_specificity)
+        ensure_specificity(incremented, increment_specificity)
         incremented[-1] += 1
 
         incremented_version = Version(".".join(map(str, incremented)))
@@ -117,13 +117,13 @@ def compute_incremented_version(
     if version_base.release > max_published.release[:specificity]:
         # For example, version_base=1.2, max_published=1.1.0.4, return 1.2.0.0
         incremented = list(version_base.release)
-        extend_to_specificity(incremented, increment_specificity)
+        ensure_specificity(incremented, increment_specificity)
 
     else:
         assert version_base.release == max_published.release[:specificity]
         # For example, version_base=1.1, max_published=1.1.0.4, return 1.1.0.5
         incremented = list(max_published.release)
-        extend_to_specificity(incremented, increment_specificity)
+        ensure_specificity(incremented, increment_specificity)
         incremented[-1] += 1
 
     incremented_version = Version(".".join(map(str, incremented)))
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -85,7 +85,7 @@ def compute_incremented_version(
     else:
         compatible = SpecifierSet(f"=={version_spec}.*")
 
-    # See what the base version and specificity we used in METADATA.toml
+    # Look up the base version and specificity in METADATA.toml.
     version_base = Version(version_spec.removesuffix(".*"))
     specificity = len(version_base.release)
 
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -85,6 +85,7 @@ def compute_incremented_version(
     else:
         compatible = SpecifierSet(f"=={version_spec}.*")
 
+    # See what the base version and specificity we used in METADATA.toml
     version_base = Version(version_spec.removesuffix(".*"))
     specificity = len(version_base.release)
 
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -65,14 +65,14 @@ def extend_to_specificity(ver: list[int], specificity: int) -> None:
 
 
 def compute_incremented_version(
-    version_spec: str, published_versions: Union[list[str], list[Version]]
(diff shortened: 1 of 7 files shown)


stub_uploader/get_version.py
```diff
@@ -83,28 +60,90 @@ def check_exists(distribution: str) -> bool:
...
-
-
-if __name__ == "__main__":
-    parser = argparse.ArgumentParser()
-    parser.add_argument("typeshed_dir", help="Path to typeshed checkout directory")
-    parser.add_argument("distribution", help="Third-party distribution to build")
-    parser.add_argument(
-        "version", nargs="?", help="Base version for which to get increment"
+def extend_to_specificity(ver: list[int], specificity: int) -> None:
+    ver.extend([0] * (specificity - len(ver)))
```
srittau, line 64: Maybe add an assertion?

```suggestion
def extend_to_specificity(ver: list[int], specificity: int) -> None:
    assert specificity >= len(ver)
    ver.extend([0] * (specificity - len(ver)))
```

stub_uploader/get_version.py
```diff
@@ -83,28 +60,90 @@ def check_exists(distribution: str) -> bool:
...
+
+
+def compute_incremented_version(
+    version_spec: str, published_versions: Union[list[str], list[Version]]
```
srittau, line 68: Why allow `list[str]`? As far as I can see, it's only used in the tests. In that case, I'd prefer a utility function in the tests and not complicate the implementation. Especially since the tests don't check the implementation code path.

stub_uploader/get_version.py
```diff
@@ -83,28 +60,90 @@ def check_exists(distribution: str) -> bool:
...
+    # The second thing we try to do (but don't guarantee), is that the incremented
+    # version will satisfy the version_spec (defined precisely by the `compatible`
+    # specifier below). This allows users to have expectations of what a stub package
+    # will contain based on the upstream version they're targeting.
+    if version_spec.endswith(".*"):
+        compatible = SpecifierSet(f"=={version_spec}")
+    else:
+        compatible = SpecifierSet(f"=={version_spec}.*")
+
+    # See what the base version and specificity we used in METADATA.toml
```
srittau, line 88:
```suggestion
    # Look up the base version and specificity in METADATA.toml.
```

stub_uploader/get_version.py
```diff
@@ -83,28 +60,90 @@ def check_exists(distribution: str) -> bool:
...
+        increment_specificity = max(increment_specificity, len(max_published.release))
+        incremented = list(max_published.release)
+        extend_to_specificity(incremented, increment_specificity)
+        incremented[-1] += 1
+
+        incremented_version = Version(".".join(map(str, incremented)))
+        assert incremented_version > max_published
+        # But can't keep versioning compatible with upstream...
+        assert incremented_version not in compatible
+        return incremented_version
```
srittau, line 115: I know that this block is basically the point of the PR, *but* could we just raise an exception here for now and wait one or two weeks before adding it in a second PR? I just want to make sure that the other changes break anything before making the functional changes. (This also makes this PR a pure refactoring.)

stub_uploader/get_version.py
```diff
@@ -83,28 +60,90 @@ def check_exists(distribution: str) -> bool:
...
+    # See what the base version and specificity we used in METADATA.toml
+    version_base = Version(version_spec.removesuffix(".*"))
+    specificity = len(version_base.release)
+
+    if max_published.epoch > 0 or version_base.epoch > 0:
+        raise NotImplementedError("Epochs in versions are not supported")
+
+    # We'll try to bump the fourth part of the release. So e.g. if our version_spec is
+    # 1.1, we'll release 1.1.0.0, 1.1.0.1, 1.1.0.2, ...
+    # But if our version_spec is 5.6.7.8, we'll release 5.6.7.8.0, 5.6.7.8.1, ...
```
srittau, line 97: Why change this behavior? For example, types-tzlocal currently has version 4.2.2, while tzlocal has version 4.2.

stub_uploader/get_version.py
```diff
@@ -83,28 +60,90 @@ def check_exists(distribution: str) -> bool:
...
+        incremented_version = Version(".".join(map(str, incremented)))
+        assert incremented_version > max_published
+        assert incremented_version in compatible
+        return incremented_version
+
+    assert version_base.release == max_published.release[:specificity]
+    # For example, version_base=1.1, max_published=1.1.0.4, return 1.1.0.5
+    incremented = list(max_published.release)
+    extend_to_specificity(incremented, increment_specificity)
+    incremented[-1] += 1
```
srittau, line 131:
```suggestion
    else:
        assert version_base.release == max_published.release[:specificity]
        # For example, version_base=1.1, max_published=1.1.0.4, return 1.1.0.5
        incremented = list(max_published.release)
        extend_to_specificity(incremented, increment_specificity)
        incremented[-1] += 1
```

srittau (dismissed):

hauntsaninja:
Thanks for the review!!

stub_uploader/get_version.py
hauntsaninja: Made this change

stub_uploader/get_version.py
hauntsaninja: That's not guaranteed to be the case (and this behaviour is tested). I renamed the function to try to make that clearer.

stub_uploader/get_version.py
hauntsaninja: I think it helps disambiguate "upstream version" from "types version". I've seen someone be confused by this and pin unnecessarily strictly.

E.g. if tzlocal released 4.2.1 that would still be compatible with 4.2 specified in METADATA.toml. But the correct corresponding version of types-tzlocal is still 4.2.2.

The issue comes about because the target version in METADATA.toml is hidden information to the average user. Very few packages use the fourth spot, so defaulting to the fourth spot in practice makes it slightly clearer that the version isn't 1:1 with upstream versions.

JelleZijlstra (approved):

stub_uploader/get_version.py
srittau: I'm fine, but I'm concerned this could break types-tzlocal (and potentially others). I would like to see this changed in a separate PR to keep possible breakage per PR minimal.

stub_uploader/get_version.py
srittau: This still needs addressing.

stub_uploader/get_version.py
hauntsaninja: I'm curious if your concern can be turned into suggestions for tests. I just added a `test_version_increment` that runs `determine_incremented_version` on all distributions we publish. This checks that all invariants we assert are met for all currently distributed packages.

But that aside, I can comment this change out for now.

hauntsaninja:
Bump, I think I've made all requested changes + added more tests

srittau:
I'm sorry, I have no time to review, but please merge if you and @JelleZijlstra are satisfied. And thanks for working on this!

JelleZijlstra (approved):

JelleZijlstra:
Merging since I'm not sure @hauntsaninja has merge access here.

PR #58, opened 2022-09-14
Move around some "types-" prefix logic

hauntsaninja:
As we plan to allow non types- dependencies, it becomes more important to be clear about when we're referring to types- packages and when not. Automatically prefixing types- in `fetch_pypi_versions` was a little surprising to me.


srittau (approved):
Good change!

hauntsaninja:
I don't have permissions on this repo, one of you or @JelleZijlstra will need to merge

JelleZijlstra:
I may not have time to review, @srittau feel free to merge if you are happy with the change.

PR #59, opened 2022-09-16
Add "requires_typeshed" property to prepare for external dependencies

hauntsaninja:
~See https://github.com/python/typeshed/issues/8312 (and https://github.com/python/typeshed/issues/5768 )~

### DIFF
--- a/data/test_typeshed/stubs/multifilepkg/METADATA.toml
+++ b/data/test_typeshed/stubs/multifilepkg/METADATA.toml
@@ -1,2 +1,2 @@
 version = "0.1"
-requires_typeshed = []
+requires = []
--- a/data/test_typeshed/stubs/multifilepkg/METADATA.toml
+++ b/data/test_typeshed/stubs/multifilepkg/METADATA.toml
@@ -1,2 +1,2 @@
 version = "0.1"
-requires = []
+requires_typeshed = []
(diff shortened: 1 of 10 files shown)


ilevkivskyi:
Not allowing runtime packages was intentional, otherwise this becomes a security hole. (Especially taking into account that the plan was to give more people typeshed commit rights in future).

ilevkivskyi:
cc @JukkaL

hauntsaninja:
This is a feature we absolutely need, so we should discuss how we can make it happen. Maybe we can have an allowlist of packages like numpy and cryptography that we trust? Let's move discussion to https://github.com/python/typeshed/issues/5768 

I'll also remove `requires_external` from this PR to reduce scope while we figure out what is needed to make it okay.

hauntsaninja:
@srittau I've updated this PR to not make any changes to METADATA.toml (or to allow external dependencies). This PR now basically just updates all existing uses of `requires` that assume it contains typeshed dists to use `requires_typeshed` property. We also tweak some exceptions and get `packaging` to do some more stuff for us.

hauntsaninja:
Actually, I'm going to close this out and fold it into the PR that allows external dependencies. This PR ended up pretty watered down and some of the code that this touches goes away, so not sure that it's that helpful to have this split.

PR #60, opened 2022-09-17
Finish changing versioning scheme

hauntsaninja:
This change was removed from #57, so as to make that PR more a refactoring. The changes here accomplish two things:

First, allows us to change the specificity of how we pin versions in typeshed while ensuring that users who install without pinning versions still get the latest stub.

Second, incrementing the fourth position helps disambiguate "upstream version" from "types version", reducing user confusion. See my comment
here: https://github.com/typeshed-internal/stub_uploader/pull/57#discussion_r956395229

We can wait longer if we're suspicious of #57, but there have been a half dozen stub uploads since then. Please let me know if there are any tests I could add that would be worthwhile.

### DIFF
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -89,16 +89,12 @@ def compute_incremented_version(
     if max_published.epoch > 0 or version_base.epoch > 0:
         raise NotImplementedError("Epochs in versions are not supported")
 
-    increment_specificity = specificity + 1
-    # TODO: uncomment this in follow-up PR
     # We'll try to bump the fourth part of the release. So e.g. if our version_spec is
     # 1.1, we'll release 1.1.0.0, 1.1.0.1, 1.1.0.2, ...
     # But if our version_spec is 5.6.7.8, we'll release 5.6.7.8.0, 5.6.7.8.1, ...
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

hauntsaninja:
I don't have permissions on this repo to merge, so one of you two would need to do it :-)

We test that all of the asserts hold when incrementing all currently existing package versions here: https://github.com/typeshed-internal/stub_uploader/blob/4588d577d7d1af180d2f0794437d1c1e78afeac3/tests/test_integration.py#L37

I also just added:
```diff
diff --git a/stub_uploader/get_version.py b/stub_uploader/get_version.py
index 9c5bfd1..c2c4bd7 100644
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -139,4 +139,5 @@ def determine_incremented_version(metadata: Metadata) -> str:
     version = compute_incremented_version(
         metadata.version_spec, published_stub_versions
     )
+    print(f"{metadata.stub_distribution:<40}\t{str(version):<12} {str(metadata.version_spec):<10} {max(published_stub_versions)}")
     return str(version)
```
and ran `pytest -s -k test_version_increment` and the results all look expected to me

PR #61, opened 2022-09-22
Allow external dependencies

hauntsaninja:
Resolves https://github.com/python/typeshed/issues/5768 (on stub_uploader side)

In https://github.com/python/typeshed/issues/5768#issuecomment-1251375258 consensus was seemingly reached on Akuli's idea, that external dependencies for stub packages are fine, as long as we validate that the external dependency is a dependency of the upstream package.

The most important part of the implementation is the validation we perform. This happens in `verify_typeshed_req` and `verify_external_req`. To help lock things down, Metadata does not expose access to elements of `requires` without validation.

We rely on PyPI's API to find the dependencies of the upstream. I believe this might not work if our stub has external dependencies and the upstream does not publish wheels. This is not the case currently (as proven by test), does not seem too likely, and it's unclear how to do safely, so will leave as a TODO for the future.

We use uploaded_packages.txt as the source of truth for what is a typeshed dependency. This is important to avoid potential badness around addition and deletion, for instance, if something is added to typeshed, but the distribution is created by someone else before stub_uploader uploads it.

The other set of changes is that I delete most of the graph code that existed previously. The graph code was added in
https://github.com/typeshed-internal/stub_uploader/pull/1 and was previously load bearing for security. The idea being to ensure that the graph of transitive dependencies was fully contained within typeshed. This is no longer the case, so we can remove most of it.

I still have some graph code in here, but it's no longer load bearing for security. I keep it around to better preserve uploading semantics, since it seems like it could matter in some edge case scenarios (such as multiple packages being uploaded for the first time that depend on each other). Since we don't have custom needs, we can get away with using the new-ish graphlib from stdlib.

While the graph code has been removed, note that we do still run validation on transitive dependencies for each package. This is accomplished by `recursive_verify`. (I think the non-transitive validation is probably sufficient, but running this before uploading each package can't hurt)

I added some special-casing for types-gdb. As Akuli pointed out, this avoids us accidentally trusting a gdb package on PyPI if ever someone attempts to add external dependencies to types-gdb.

New code paths have tests. I audited test coverage to make sure of this.

### DIFF
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -135,7 +135,7 @@ uploaded_packages = _UploadedPackages(UPLOADED_PATH)
 def strip_types_prefix(dependency: str) -> str:
     if not dependency.startswith(TYPES_PREFIX):
         raise ValueError("Expected dependency on a typeshed package")
-    return dependency[len(TYPES_PREFIX) :]
+    return dependency.removeprefix(TYPES_PREFIX)
 
 
 def verify_typeshed_req(req: Requirement) -> None:
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -49,10 +49,9 @@ class Metadata:
 
     @property
     def _unvalidated_requires_typeshed(self) -> list[Requirement]:
+        typeshed = uploaded_packages.read()
         return [
-            r
-            for r in self._unvalidated_requires
-            if canonical_name(r.name) in uploaded_packages.read()
+            r for r in self._unvalidated_requires if canonical_name(r.name) in typeshed
         ]
 
     @functools.cached_property
@@ -64,10 +63,11 @@ class Metadata:
 
     @property
     def _unvalidated_requires_external(self) -> list[Requirement]:
+        typeshed = uploaded_packages.read()
         return [
             r
             for r in self._unvalidated_requires
-            if canonical_name(r.name) not in uploaded_packages.read()
+            if canonical_name(r.name) not in typeshed
         ]
 
     @functools.cached_property
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -159,6 +159,7 @@ def verify_typeshed_req(req: Requirement) -> None:
 EXTERNAL_REQ_ALLOWLIST = {
     "numpy",
     "cryptography",
+    "torch",
 }
 
 
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -193,6 +193,9 @@ def verify_external_req(
     # TODO: consider allowing external dependencies for stubs for packages that do not ship wheels.
     # Note that we can't build packages from sdists, since that can execute arbitrary code.
     # We could do some hacky setup.py parsing though...
+    # TODO: PyPI doesn't seem to have version specific requires_dist. This does mean we can be
+    # broken by new releases of upstream packages, even if they do not match the version spec we
+    # have for the upstream distribution.
     if req.name not in [
         Requirement(r).name for r in (data["info"].get("requires_dist") or [])
     ]:
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -159,8 +159,9 @@ def verify_external_req(req: Requirement, upstream_distribution: Optional[str]) 
         )
 
     if upstream_distribution is None:
-        # There is not an upstream distribution on PyPI
-        return
+        raise InvalidRequires(
+            f"There is no upstream distribution on PyPI, so cannot verify {req}"
+        )
 
     resp = requests.get(f"https://pypi.org/pypi/{upstream_distribution}/json")
     if resp.status_code != 200:
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -1,19 +1,39 @@
+from __future__ import annotations
+
+import functools
+import graphlib
 import os
-from typing import Any, Dict, List, Optional
+import re
+from typing import Any, Dict, Iterator, Optional
 
+import requests
 import tomli
+from packaging.requirements import Requirement
 
-from .const import META, THIRD_PARTY_NAMESPACE, TYPES_PREFIX
+from .const import META, THIRD_PARTY_NAMESPACE, TYPES_PREFIX, UPLOADED_PATH
+
+
+class InvalidRequires(Exception):
+    pass
 
 
 class Metadata:
     def __init__(self, distribution: str, data: Dict[str, Any]):
-        self.upstream_distribution = distribution
+        assert not distribution.startswith(TYPES_PREFIX)
+        self._alleged_upstream_distribution = distribution
         self.data = data
 
     @property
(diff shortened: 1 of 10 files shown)


stub_uploader/metadata.py
```diff
@@ -42,7 +92,135 @@ def no_longer_updated(self) -> bool:
...
+            f"Expected dependency {req} to not be uploaded from stub_uploader"
+        )
+    if req.name.startswith(TYPES_PREFIX):
+        # technically this could be allowed, but it's very suspicious
+        raise InvalidRequires(
+            f"Expected dependency {req} to not start with {TYPES_PREFIX}"
+        )
+
+    if upstream_distribution is None:
+        # There is not an upstream distribution on PyPI
```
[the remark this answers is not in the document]
hauntsaninja: Good spot, fixed! This is what I get for adding types-gdb hackiness at the last minute

stub_uploader/metadata.py
```diff
@@ -42,7 +92,135 @@ def no_longer_updated(self) -> bool:
...
+
+
+class _UploadedPackages:
+    def __init__(self, file_path: str) -> None:
+        self._file_path = file_path
+        self._cached: Optional[list[str]] = None
+
+    def read(self) -> set[str]:
+        if self._cached is not None:
+            return set(map(canonical_name, self._cached))
```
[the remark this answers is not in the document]
hauntsaninja: Currently uploaded_packages.txt uses pretty case and I figured we might want to keep that

stub_uploader/metadata.py
```diff
@@ -24,8 +44,38 @@ def version_spec(self) -> str:
...
+        for req in reqs:
+            verify_typeshed_req(req)
+        return reqs
+
+    @property
+    def _unvalidated_requires_external(self) -> list[Requirement]:
+        return [
+            r
+            for r in self._unvalidated_requires
+            if canonical_name(r.name) not in uploaded_packages.read()
```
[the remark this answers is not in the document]
hauntsaninja: Thanks, fixed

stub_uploader/metadata.py
```diff
@@ -1,19 +1,39 @@
...
-        self.upstream_distribution = distribution
+        assert not distribution.startswith(TYPES_PREFIX)
+        self._alleged_upstream_distribution = distribution
         self.data = data
 
+    @property
+    def upstream_distribution(self) -> Optional[str]:
+        # TODO: add a field to METADATA.toml if the stubs are for a package
+        # that does not exist on PyPI
+        if self._alleged_upstream_distribution == "gdb":
```
ilevkivskyi, line 30: This makes me think about another possible attack. What if someone creates a package `foo` and adds `types-foo` to typeshed. It can be a totally benign package. Then later they make a PR that makes `types-foo` a dependency of something commonly used, like `types-requests` (or maybe something less suspicious), e.g. because the library has some custom type that works as an argument for some `requests` function. Much later they add some malicious package `bad` as a dependency of `foo`, and submit a "trivial" PR adding it also as a dependency of `types-foo`. Now this malicious package `bad` will be installed for _a lot_ of users (executing arbitrary code during install), despite the fact that no-one may actually use `foo`.

Of course this is a tricky scenario, but someone may still try it, taking into account that potential impact will be very high. You may say that something similar is also possible for a regular runtime package, but the difference is that maintainers of runtime packages usually deal with only few dependencies, and adding another one is a really big deal, while typeshed maintainers will (hopefully) deal with hundreds or thousands packages each with their own deps.

So, I would say we will still need an explicit allowlist _in this repo_, in addition to the logic proposed in this PR.

hauntsaninja: I think that can be dealt with in a similar way to the current logic for external dependencies, that is, in order to add `types-xyz` as a dependency of `types-requests`, `xyz` needs to be a dependency of `requests`.

That's probably more robust too, because whatever process we have for adding things to the allowlist will be used to add things to the allowlist. To spell things out:
1. Package foo exists, that depends on bar
2. Add types-foo, add dependency on bar
3. In order to make types-foo useful, we need bar, so we do some security review of bar and add it to the allowlist
4. bar adds dependency on evil
5. types-foo users are affected, this isn't good but we're okay with it, since those are assumed to be a subset of foo users
6. Someone sneaks past adding a dependency on types-foo to types-requests
7. types-requests users are affected, this is really bad

So adding a step 3 can help in practice, but we should prevent step 6. I can write this check, should be pretty easy to extend the existing logic.

Note that I'm not against adding an allowlist, but I'd like to think of it more as a fallback. I'd also want us to clearly document the criteria for additions to the allowlist.

hauntsaninja: Okay, I wrote the logic for preventing step 6 locally, but it doesn't pass cleanly against existing packages. There are four failures, two of which are addressed by https://github.com/python/typeshed/pull/8792 and https://github.com/python/typeshed/pull/8793.

The other two are trickier, so would need to make that change later in another PR (might need new METADATA.toml fields). In the meantime I've added the allowlist as proposed, which should unblock this PR.

stub_uploader/metadata.py
```diff
@@ -141,13 +141,32 @@ def strip_types_prefix(dependency: str) -> str:
...
+    # TODO: make sure that if a typeshed distribution depends on other typeshed stubs,
+    # the upstream depends on the upstreams corresponding to those stubs.
+    # See https://github.com/typeshed-internal/stub_uploader/pull/61#discussion_r979327370
+
+
+# Presence in the top 1000 PyPI packages could be a necessary but not sufficient criterion for
+# inclusion in this allowlist.
+# Note we could loosen our criteria once we address:
+# https://github.com/typeshed-internal/stub_uploader/pull/61#discussion_r979327370
+EXTERNAL_REQ_ALLOWLIST = {
```
AlexWaygood, line 159: `pytorch` would also be useful for at least one package, off the top of my head

```suggestion
EXTERNAL_REQ_ALLOWLIST = {
    "pytorch",
```

hauntsaninja: Sure, although note that the dist name is torch ;-)

stub_uploader/metadata.py
```diff
@@ -42,7 +92,165 @@ def no_longer_updated(self) -> bool:
...
+    "numpy",
+    "cryptography",
+}
```
hauntsaninja, line 162:
```suggestion
    "cryptography",
    "torch",
}
```

stub_uploader/metadata.py
Avasam: D3DShot uses torch.

PyInstaller could be simplified by having access to modulegraph.

stub_uploader/metadata.py
ilevkivskyi: > but it doesn't pass cleanly against existing packages

Yeah, I felt it may happen. Anyway, the PR looks good now.

ilevkivskyi (approved):

hauntsaninja:
Bump :-)

ilevkivskyi:
@hauntsaninja Hm, it looks like one of added tests is broken on master, could you please check?

PR #69, opened 2022-10-14
Extract upload() from upload_changed/some

srittau:
Both code segments were essentially identical. This is a first step to identifying why our uploads fails.


srittau:
@ilevkivskyi @JelleZijlstra @hauntsaninja @alexwaygood @Akuli Sorry for the spam, but could one of you review this so we can get this resolved ASAP?

PR #76, opened 2022-11-22
Support a "stub_distribution" metadata field

srittau:
This allows us to override the generated and uploaded distribution name,
e.g. if the name is already in use.

Cf. python/typeshed#9246


srittau:
Closing in favor of #79.

PR #78, opened 2022-11-22
Refactor stub dist name in templates

srittau:
Use Metadata.stub_distribution instead of constructing the name
manually in string templates. This ensures consistency and is
safer if we change the distribution name format.


srittau:
Split from #76, since I think this is independently useful.

JelleZijlstra (approved):

PR #79, opened 2022-11-23
Add "stub_distribution" and "upload" metadata fields

srittau:
These optional fields allow more control over what gets uploaded from the typeshed side:

* `stub_distribution` allows to change the uploaded distribution name.
* Setting `upload` to `false` prevents uploading a package (e.g. in case of errors).

### DIFF
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -4,8 +4,8 @@ import functools
 import graphlib
 import os
 import re
-from typing import Any, Optional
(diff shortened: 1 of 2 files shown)


srittau:
I believe that these fields could prove useful in the future, not matter what we decide to do for python/typeshed#9246.

AlexWaygood (approved):
Nice solution.

ilevkivskyi (approved):

PR #83, opened 2023-01-04
Fix dependency_order bug

hauntsaninja:
Unblocks https://github.com/python/typeshed/pull/9459


JelleZijlstra (approved):

AlexWaygood (approved):
This makes sense to me (though I'm pretty new to this codebase)

PR #84, opened 2023-01-06
Mention pyright in the stubs' README

Avasam:
Same suggestion as https://github.com/python/typeshed/pull/9472

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -2,12 +2,13 @@
 Basic script to generate a wheel for a third-party distribution in typeshed.
 
 This generates a PEP 561 types stub package using METADATA.toml file for a given
-distribution in typeshed stubs. Such package can be used by type-checking tools
+distribution in typeshed stubs. Such package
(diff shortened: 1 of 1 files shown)


AlexWaygood (commented):
Maybe add [`pyre`](https://github.com/facebook/pyre-check) to this list as well? `pyre` has 6.2k stars on GitHub compared to e.g. 4k stars for `pytype` :)

srittau (approved):

Avasam:
> Maybe add [`pyre`](https://github.com/facebook/pyre-check) to this list as well? `pyre` has 6.2k stars on GitHub compared to e.g. 4k stars for `pytype` :)

Sure can, I just didn't remember about pyre since typeshed doesn't text directly against it.

srittau:
I think we should skip pyre for now, since the stubs are not tested against it.

AlexWaygood:
> I think we should skip pyre for now, since the stubs are not tested against it.

No strong opinion! This PR is a clear improvement as-is; don't want to block it

PR #87, opened 2023-01-14
Case insensitive check between specified requirement and requires_dist

### LINKED ISSUE #86: Dependency check is case-sensitive
This is problematic for packages like JACK-Client that declared its `numpy` dependency as `NumPy`. Since PyPI packages are case-insensitive, I think the check against requires_dist should be as well.

Ref: https://github.com/python/typeshed/pull/9530

Avasam:
Attempts to fix #86

### DIFF
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -209,7 +209,8 @@ def verify_external_req(
     # broken by new releases of upstream packages, even if they do not match the version spec we
     # have for the upstream distribution.
     if canonical_name(req.name) not in [
-        canonical_name(Requirement(r).name) for r in (data["info"].get("requires_dist") or [])
+        canonical_name(Requirement(r).name)
+        for r in (data["info"].get("requires_dist") or [])
     ]:
         raise InvalidRequires(
             f"Expected dependency {req} to be listed in {upstream_distribution}'s requires_dist"
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -179,7 +179,9 @@ def verify_external_req(
     upstream_distribution: Optional[str],
     _unsafe_ignore_allowlist: bool = False,  # used for tests
 ) -> None:
-    if canonical_name(req.name) in uploaded_packages.read():
+    req_canonical_name = canonical_name(req.name)
+
+    if req_canonical_name in uploaded_packages.read():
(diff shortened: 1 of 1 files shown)


hauntsaninja (changes requested):
Thanks! Could you use canonical_name instead?

stub_uploader/metadata.py
```diff
@@ -208,8 +208,8 @@ def verify_external_req(
     # TODO: PyPI doesn't seem to have version specific requires_dist. This does mean we can be
     # broken by new releases of upstream packages, even if they do not match the version spec we
     # have for the upstream distribution.
-    if req.name not in [
-        Requirement(r).name for r in (data["info"].get("requires_dist") or [])
+    if req.name.lower() not in [
+        Requirement(r).name.lower() for r in (data["info"].get("requires_dist") or [])
```
Avasam, line 212: Can I assume that the requirement we manually specify is always a canonical name? Or should I normalize both like this?
```suggestion
    if canonical_name(req.name) not in [
        canonical_name(Requirement(r).name) for r in (data["info"].get("requires_dist") or [])
```

stub_uploader/metadata.py
```diff
@@ -208,8 +208,9 @@ def verify_external_req(
     # TODO: PyPI doesn't seem to have version specific requires_dist. This does mean we can be
     # broken by new releases of upstream packages, even if they do not match the version spec we
     # have for the upstream distribution.
-    if req.name not in [
-        Requirement(r).name for r in (data["info"].get("requires_dist") or [])
+    if canonical_name(req.name) not in [
+        canonical_name(Requirement(r).name)
+        for r in data["info"].get("requires_dist", [])
```
Avasam, line 213:
```suggestion
        for r in (data["info"].get("requires_dist") or [])
```

stub_uploader/metadata.py
JelleZijlstra: I guess either we should normalize both or we should test somewhere (probably typeshed) that all reqs are in canonical form.

Avasam: Since it was being called twice (thrice with #88). I figured may as well deduplicate it. And now it's only called once anyway, so may as well use the canonical name.

srittau (approved):

PR #88, opened 2023-01-15
get sdist requires from egg-info

Avasam:
CC @AlexWaygood This was the idea I mentioned.

When the requirements can't be found in `requires_dist`, fallback to searching in *.egg-info. All without executing any arbitrary code.

I move the `f"Expected dependency {req} to be present in the allowlist"` error higher so it can fail earlier and leave the network requests last.

It's my first time playing around with downloading and manipulating archives in python. Let me know what I can do better here.

I tested the idea with `pycocotools` (best case scenario) `click-default-group` (on my windows machine with 7zip there's an extra folder between the gz and tar, still worked), `numpy` (no *.egg-info folder), `mypy` (has empty lines and [extras]) and `pywin32` (no Source Distribution)


Avasam:
Tests found at least one case where this errors out. I'll look into it.

stub_uploader/metadata.py
```diff
@@ -190,30 +235,27 @@ def verify_external_req(
...
+    if req.name in [
         Requirement(r).name for r in (data["info"].get("requires_dist") or [])
     ]:
-        raise InvalidRequires(
-            f"Expected dependency {req} to be listed in {upstream_distribution}'s requires_dist"
-        )
+        return  # Ok!
 
-    if req.name not in EXTERNAL_REQ_ALLOWLIST and not _unsafe_ignore_allowlist:
+    if req_canonical_name not in get_sdist_requires_canonical(data["urls"][-1], req):
```
Avasam, line 255: > `data["urls"][-1]`

Two assumptions here:
1. The Source Distribution is always last, Built Distributions (wheels) will come first
2. If there's no Source Distribution (like `pywin32`, this will incidentally take a wheel. But if wheels are available we shouldn't get here in the first place.

I think it's worth checking for `python_version == "source"` to be more rigorous.

Avasam:
Should I delete the files afterward or does it not matter as code meant for the CI?

stub_uploader/metadata.py
```diff
@@ -174,6 +177,44 @@ def verify_typeshed_req(req: Requirement) -> None:
...
+    matches = glob(str(requires_filepath))
+    for match in matches:
+        lines: list[str] = []
+        with open(match) as requires_file:
+            lines = requires_file.readlines()
+        for line in lines:
+            # Skip empty lines and extras
+            if line[0] not in {"\n", "["}:
+                yield Requirement(line)
+        # Assume a single *.egg-info folder
```
Avasam, line 214: Same with the `# Assume a single folder` comment above. These are assumptions from what I saw testing a few distributions.
I could also check the length, give an error if 2 or more (then we can investigate and correctly deal with those cases).
And stop early if 0. Not sure what's the best practice for returning an empty Generator. `return`, `return iter(())`, `yield from ()` ?

stub_uploader/metadata.py
```diff
@@ -174,6 +177,43 @@ def verify_typeshed_req(req: Requirement) -> None:
...
+    with open(filename, "wb") as file:
+        file.write(resp.raw.read())
+
+    folder_name = ""
+    with tarfile.open(filename) as file_in:
+        file_in.extractall()
+        for info in file_in:
+            folder_name = info.name
+            # Assume a single folder
+            break
```
srittau, line 202: I would add a check here, whether `folder_name` was actually set, or maybe even check that there really is exactly one folder to account for broken eggs. But for the former case, the following should be simpler and have that required error check:

```suggestion
        # Assume a single folder
        folder_name = next(iter(file_in)).name
```

You can then also drop the initialization of `folder_name`.

Or just use `*` instead of `folder_name` in the glob below.

stub_uploader/metadata.py
```diff
@@ -174,6 +177,43 @@ def verify_typeshed_req(req: Requirement) -> None:
...
 
 
+def validate_response(resp: requests.Response, req: Requirement) -> None:
+    if resp.status_code != 200:
+        raise InvalidRequires(
+            f"Expected dependency {req} to be accessible on PyPI, but got {resp.status_code}"
+        )
+
+
+def get_sdist_requires(
```
srittau, line 187: Nit: "get" makes this function sound simpler than what it actually does. Maybe "extract"?

Also I think it would be safer to extract the sdist into an empty temporary directory than to where ever we currently are.

srittau (changes requested):
Thanks, remarks below.

AlexWaygood:
At least some of the test failures will be fixed by https://github.com/typeshed-internal/stub_uploader/pull/95

Avasam:
I was thinking integration tests might be too fragile, but I can just pin the version I'm testing against. I'm sure you'd like me to add  tests for this? 

Quick testing script I've been using:
```py
from __future__ import annotations

import requests
from packaging.requirements import Requirement

from stub_uploader.metadata import extract_sdist_requires, validate_response

for req_name in ("pycocotools", "click-default-group", "numpy", "mypy", "pywin32"):
    req = Requirement(req_name)

    resp = requests.get(f"https://pypi.org/pypi/{req.name}/json")
    validate_response(resp, req)
    data = resp.json()

    sdist_data = next(
        (
            url_data
            for url_data in reversed(data["urls"])
            if url_data["packagetype"] == "sdist"
        ),
        None,
    )
    if sdist_data is None:
        print(f"{req_name}: sdist_data is None!")
        continue

    requirements = extract_sdist_requires(sdist_data, req)
    print(f"{req_name}: {list(requirements)}")
```

Results:
```py
pycocotools: [<Requirement('matplotlib>=2.1.0')>, <Requirement('numpy')>]
click-default-group: [<Requirement('click')>]
numpy: []
mypy: [<Requirement('typing_extensions>=3.10')>, <Requirement('mypy_extensions>=1.0.0')>, <Requirement('tomli>=1.1.0')>, <Requirement('typed_ast<2,>=1.4.0')>, <Requirement('psutil>=4.0')>, <Requirement('pip')>, <Requirement('typed_ast<2,>=1.4.0')>, <Requirement('lxml')>]
pywin32: sdist_data is None!
```

srittau (approved):

PR #94, opened 2023-05-02
Allow marking stubs as partial

### LINKED ISSUE #93: Mark incomplete stubs as `partial`
Incomplete stubs (`ignore_missing_stub = true`) should probably have `partial\n` in a `py.typed` marker. Ref: https://peps.python.org/pep-0561/#partial-stub-packages

Avasam:
Closes #93

I'm sure this can be improved. Also please review and test carefully (I'm still running integration tests, but they're taking forever).

~~The basic idea is to infer a stub package being `partial` from finding `ignore_missing_stub = true` in the metadata. If we decide to add a specific "partial" entry in the metadata, the logic will stay the same, only the property's inner code will change.~~
Edit: went for a separate metadata key.

Alternatively typeshed could add the `partial\n` `py.typed` marker itself (assuming it can use packages instead of single files for single-file source w/o problem with stubtest)

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -273,11 +273,7 @@ def generate_long_description(
             )
         )
     elif metadata.partial:
-        parts.append(
-            NO_LONGER_UPDATED_TEMPLATE.format(
-                stub_distribution=metadata.stub_distribution
-            )
-        )
+        parts.append(PARTIAL_STUBS_DESCRIPTION)
     parts.append(DESCRIPTION_OUTRO_TEMPLATE.format(commit=commit))
     return "\n\n".join(parts)
 
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -38,6 +38,11 @@ CHANGELOG = "CHANGELOG.md"
 
 SUFFIX = "-stubs"
 
+PARTIAL_STUBS_DESCRIPTION = """
+This stub package is marked as [partial](https://peps.python.org/pep-0561/#partial-stub-packages).
+If you find that annotations are missing, feel free to contribute and help complete them.
+""".lstrip()
+
 SETUP_TEMPLATE = dedent(
     """
 from setuptools import setup
@@ -110,7 +115,7 @@ def find_stub_files(top: str) -> list[str]:
 
     Raise if we find any unknown file extensions during collection.
     """
-    result = []
+    result: list[str] = []
     for root, _, files in os.walk(top):
         for file in files:
             if file.endswith(".pyi"):
@@ -140,7 +145,8 @@ def copy_stubs(base_dir: str, dst: str) -> None:
             if not entry.endswith(".pyi"):
                 continue
             stub_dir = os.path.join(dst, entry.split(".")[0] + SUFFIX)
-            os.mkdir(stub_dir)
+            if not os.path.exists(stub_dir):
+                os.mkdir(stub_dir)
             shutil.copy(
                 os.path.join(base_dir, entry), os.path.join(stub_dir, "__init__.pyi")
             )
@@ -180,7 +186,7 @@ def collect_setup_entries(base_dir: str) -> dict[str, list[str]]:
 
     This reflects the transformations done during copying in copy_stubs().
     """
-    package_data = {}
+    package_data: dict[str, list[str]] = {}
     for entry in os.listdir(base_dir):
         if entry == META:
             # Metadata file entry is added at the end.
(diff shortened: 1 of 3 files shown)


srittau:
In that case it would probably make sense to rename `ignore_missing_stub` to `partial` or `partial_stub` and clearly document what it does on the typeshed side. I'm a bit uncomfortable about adding the partial marker just by looking at some field named `ignore_missing_stub`.

Avasam:
> In that case it would probably make sense to rename `ignore_missing_stub` to `partial` or `partial_stub` and clearly document what it does on the typeshed side. I'm a bit uncomfortable about adding the partial marker just by looking at some field named `ignore_missing_stub`.

I personally wouldn't rename `[tool.stubtest].ignore_missing_stub` (at least not for this feature).

I infer from it, because having stubtest ignore missing stubs usually means that, well, there's missing stubs! (and so, the stub package should *probably* be marked as partial, so that we don't hide public API that is not yet typed).
I do recognize this may not always be the case in the future (as we both alluded to `ignore_missing_stub` and `partial` are semantically different), so if you'd prefer going immediately for a different setting, let's do that :) (and the required changes in typeshed).

I see two ways of doing it (more suggestions welcome). Whichever solution should be well documented in typeshed.
1. Let typeshed add `py.typed` markers itself. And a test to ensure it contains `partial\n`. stub_uploader then only needs to include them in its filters.
2. Add a new metadata entry `partial` or `partial_stub`, keep this PR mostly as-is.

Now for example use cases where `partial_stub = true` and `ignore_missing_stub = false`:
- pywin32: The maintainers stated not wanting to maintain separate stub files. But are open to inline typing (given they can be validated). So once ongoing obsolete code cleanup is done, the library would use inline types, and c-module extensions stubs would live in typeshed. I'd suppose pure-python modules would be excluded in stubtest. This is a very long-term maybe-plan.

`partial_stub = false` and `ignore_missing_stub = true`:
- stubs that are skipped in stubtest (though `skip` + `ignore_missing_stub` wouldn't hurt)
- stubs that would have stubtest_allowlist files of thousands of lines (I don't have a concrete example where this is a preferred solution over "fixing" whatever issue with a stub that causes this, or using regex)

srittau:
I'd prefer to split those features, i.e. have a separate `partial_stub` key in the `METADATA.toml` file for the use cases you outlined and to avoid "hidden" side effects.

stub_uploader/build_wheel.py
```diff
@@ -254,6 +272,12 @@ def generate_long_description(
                 stub_distribution=metadata.stub_distribution
             )
         )
+    elif metadata.partial:
+        parts.append(
+            NO_LONGER_UPDATED_TEMPLATE.format(
+                stub_distribution=metadata.stub_distribution
+            )
+        )
```
Avasam, line 280:
```suggestion
        parts.append(PARTIAL_STUBS_DESCRIPTION)
```

stub_uploader/build_wheel.py
```diff
@@ -140,7 +145,8 @@ def copy_stubs(base_dir: str, dst: str) -> None:
             if not entry.endswith(".pyi"):
                 continue
             stub_dir = os.path.join(dst, entry.split(".")[0] + SUFFIX)
-            os.mkdir(stub_dir)
+            if not os.path.exists(stub_dir):
+                os.mkdir(stub_dir)
```
srittau, line 149:
```suggestion
            os.makedirs(stub_dir, exist_ok=True)
```

stub_uploader/build_wheel.py
```diff
@@ -211,6 +217,16 @@ def collect_setup_entries(base_dir: str) -> dict[str, list[str]]:
     return package_data
 
 
+def add_partial_marker(package_data: dict[str, list[str]], stub_dir: str) -> None:
+    for entry, files in package_data.items():
+        entry_path = os.path.join(stub_dir, entry)
+        if not os.path.exists(entry_path):
+            os.mkdir(entry_path)
```
srittau, line 224:
```suggestion
        os.makedirs(entry_path, exist_ok=True)
```

stub_uploader/build_wheel.py
```diff
@@ -254,6 +272,8 @@ def generate_long_description(
                 stub_distribution=metadata.stub_distribution
             )
         )
+    elif metadata.partial:
```
srittau, line 275: I think it might be a good idea to still have the "partial" comment, even if the package is outdated, if only for consistency reasons. Of course, it's a bit weird to ask the user to contribute to outdated packages, but I don't think it's too big of a deal.

srittau (changes requested):
Thanks! The `mkdir/makedirs` suggestions are optional, in case you like your solution better.

stub_uploader/build_wheel.py
Avasam:
```suggestion
    if metadata.partial:
```

PR #96, opened 2023-06-12
Include type checker versions in long description

srittau:
Cf. python/typeshed#6843

### DIFF
--- a/stub_uploader/ts_data.py
+++ b/stub_uploader/ts_data.py
@@ -46,9 +46,12 @@ def parse_requirements(stream: Iterable[str]) -> dict[str, str]:
         if not line.strip():  # skip empty lines
             continue
         req = Requirement(line.strip())
-        if len(req.specifier) != 1 or not str(req.specifier).startswith("=="):
+        if len(req.specifier) != 1:
             continue
-        requirements[req.name] = str(req.specifier)[2:]
+        spec = next(iter(req.specifier))
+        if spec.operator != "==":
+            continue
+        requirements[req.name] = spec.version
     return requirements
 
 
--- a/stub_uploader/ts_data.py
+++ b/stub_uploader/ts_data.py
@@ -4,10 +4,11 @@ Information about typeshed.
 
 from __future__ import annotations
 import argparse
+from collections.abc import Iterable
 from dataclasses import dataclass, fields
+from packaging.requirements import Requirement
 from pathlib import Path
 from tomli import load as toml_load
-from typing import Any
 
 
 REQUIREMENTS = "requirements-tests.txt"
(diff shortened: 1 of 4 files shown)


stub_uploader/ts_data.py
```diff
@@ -0,0 +1,63 @@
...
+def _read_requirements(path: Path) -> dict[str, str]:
+    requirements = {}
+    with path.open() as f:
+        for line in f:
+            line = line.strip()
+            line = line.split("#")[0]  # strip comments
+            line = line.split(";")[0]  # strip extras
+            if not line.strip():  # skip empty lines
+                continue
+            if "==" in line:
```
ilevkivskyi, line 48: What if at some point in future someone will (accidentally) write e.g. `mypy>=1.5.1`?

ilevkivskyi (commented):
Generally I like the idea, but I have one comment and I think it may be worth writing few unit tests for `_read_requirements()`.

srittau:
> Generally I like the idea, but I have one comment and I think it may be worth writing few unit tests for `_read_requirements()`.

Good idea, I will do so tomorrow.

stub_uploader/ts_data.py
AlexWaygood: Stub uploader already has `packaging` as a requirement, so we could use the `packaging.requirements.Requirement` class to parse each line (once it's been stripped of comments), rather than doing our own parsing. That would simplify this code and also be more robust/better tested.

srittau: I see that as a feature. Using a non-pinned version in typeshed would make tests non-reproducible. Although maybe we could reorganize requirements-tests.txt a bit and add a comment to that effect?

AlexWaygood: I agree that most of our test dependencies (and definitely our type-checker dependencies) in typeshed should be pinned. But if we want to test that, I feel like we should have a dedicated test for it, rather than having some unrelated code in stub-uploader inexplicably fail when that rule is broken.

hauntsaninja: (We run stub_uploader tests in typeshed, so if you have an integration test that makes sure this works, we should get failures at typeshed PR time)

stub_uploader/ts_data.py
```diff
@@ -0,0 +1,61 @@
...
+    requirements = {}
+    for line in stream:
+        line = line.strip()
+        line = line.split("#")[0]  # strip comments
+        if not line.strip():  # skip empty lines
+            continue
+        req = Requirement(line.strip())
+        if len(req.specifier) != 1 or not str(req.specifier).startswith("=="):
+            continue
+        requirements[req.name] = str(req.specifier)[2:]
```
AlexWaygood, line 51:
```suggestion
        if len(req.specifier) != 1:
            continue
        spec = next(iter(req.specifier))
        if spec.operator != "==":
            continue
        requirements[req.name] = spec.version
```

AlexWaygood (approved):
This looks great to me, and I think will be really useful. Thanks @srittau!

ilevkivskyi (approved):

PR #101, opened 2023-09-11
Support .postN versions

srittau:
The post version is just replaced with a regular release segment, as PEP 440 doesn't support structured post versions.

### DIFF
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -58,7 +58,7 @@ def compute_incremented_version(
     # need revisiting.
     max_published = max(published_versions, default=Version("0"))
 
-    # Parse and massage the base version. if necessary. Usually, the base version is
+    # Parse and massage the base version if necessary. Usually, the base version is
     # just the version_spec without any trailing `.*`. But if the version_spec is a
     # post version, we append the post version to the base version:
     #   1.2.post3 -> 1.2.3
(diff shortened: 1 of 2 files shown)


AlexWaygood (approved):
This seems correct from what I can tell, but I'm not particularly familiar with this code

ilevkivskyi (approved):

stub_uploader/get_version.py
```diff
@@ -62,13 +66,26 @@ def compute_incremented_version(
     # version will satisfy the version_spec (defined precisely by the `compatible`
     # specifier below). This allows users to have expectations of what a stub package
     # will contain based on the upstream version they're targeting.
-    if version_spec.endswith(".*"):
+    m = POST_SPEC_RE.match(version_spec)
+    if m:
```
hauntsaninja, line 70: We could use `Version(version_spec.removesuffix(".*")).post is not None` instead of regex parsing. E.g. I think this doesn't handle https://peps.python.org/pep-0440/#implicit-post-releases

stub_uploader/get_version.py
```diff
@@ -62,13 +66,26 @@ def compute_incremented_version(
...
+        compatible = SpecifierSet(f"=={m.group(1)}.{m.group(2)}.*")
```
hauntsaninja, line 71: Similarly, this could use `Version(version_spec.removesuffix(".*")).base_version` here

PR #105, opened 2023-10-06
Improve PyPI READMEs for stubs packages

AlexWaygood:
- Link directly to PEP-561
- Link directly to the runtime package the stubs are for, if typeshed provides the `upstream_repository` field in a stubs package's `METADATA.toml` file.
- Clearly state that `types-requests==2.31.0.8` aims to provide accurate annotations for `requests>=2.31.0`.

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -266,10 +266,7 @@ def generate_setup_file(
         distribution=build_data.distribution,
         stub_distribution=metadata.stub_distribution,
         long_description=generate_long_description(
-            build_data.distribution,
-            commit,
-            ts_data,
-            metadata,
+            build_data.distribution, commit, ts_data, metadata
         ),
         version=version,
         requires=all_requirements,
@@ -280,10 +277,7 @@ def generate_setup_file(
 
 
 def generate_long_description(
-    distribution: str,
-    commit: str,
-    ts_data: TypeshedData,
-    metadata: Metadata,
+    distribution: str, commit: str, ts_data: TypeshedData, metadata: Metadata
 ) -> str:
     extra_description = metadata.extra_description.strip()
     parts: list[str] = []
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -108,7 +108,7 @@ PyCharm, etc. to check code that uses
 `{distribution}`.
 
 This version of `{stub_distribution}` aims to provide accurate annotations
-for `{distribution_range_supported}`.
+for `{distribution}=={typeshed_version_pin}`.
 The source for this package can be found at
 https://github.com/python/typeshed/tree/main/stubs/{distribution}. All fixes for
 types and metadata should be contributed there.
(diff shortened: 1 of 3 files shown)


AlexWaygood:
We might want to switch to using jinja for some of the stuff going on in `build_wheel.py`; it can make stuff like this much more readable, in my opinion.

stub_uploader/metadata.py
```diff
@@ -112,6 +112,10 @@ def requires_python(self) -> str | None:
         verify_requires_python(req)
         return req
 
+    @property
+    def upstream_repository(self) -> str | None:
+        return self.data.get("upstream_repository")
```
AlexWaygood, line 117: Possibly we could do some validation here. Typeshed already does [a bunch of validation checks](https://github.com/python/typeshed/blob/54a8fd74a785ff7916bade3699a2be4542f1b1c0/tests/parse_metadata.py#L222-L244) for this field, though, so maybe it's not necessary.

srittau: A few general notes: Generally the stub_uploader should treat everything coming from typeshed as "tainted" as a precaution against an attacker gaining access to a typeshed maintainer's GitHub account. (This is the reason why we have so few stub_uploader maintainers -> to limit the attack vector. Personally, I consider every typeshed maintainer to be a stub_uploader maintainer, just without the commit rights.) Crashing the uploader is fine, the risk is uploading potentially harmful code under a "types-" name.

That said, an `isinstance` check and regexp with "safe" markdown chars can't hurt here. The regexp check should probably just print a warning and return `None`.

stub_uploader/build_wheel.py
```diff
@@ -272,11 +281,32 @@ def generate_setup_file(
...
-    parts.append(DESCRIPTION_INTRO_TEMPLATE.format(distribution=distribution))
+
+    if metadata.upstream_repository is not None:
+        formatted_distribution = f"[`{distribution}`]({metadata.upstream_repository})"
+    else:
+        formatted_distribution = f"`{distribution}`"
+
+    distribution_range_supported = (
+        f"{distribution}>={'.'.join(version.split('.')[:-1])}"
+    )
```
srittau, line 300: `version` is the type stubs package version, which is already processed. For example, pytz's metadata field currently specifies `2023.3.post1`, but this version is likely something like `2023.3.1.12`. Similarly, pywin32's version is `306.*`, which causes this version to be `306.0.0.5`. I guess we should just use the `version` field from the metadata file here.

AlexWaygood: Switched to using the `version` field from the metadata file in https://github.com/typeshed-internal/stub_uploader/pull/105/commits/796d2ea3bd5e9365a93c6e60287011b022a3774d

stub_uploader/metadata.py
AlexWaygood: I added a little more validation in https://github.com/typeshed-internal/stub_uploader/pull/105/commits/796d2ea3bd5e9365a93c6e60287011b022a3774d. We could add more. Should we return `None` if `.isascii()` returns `False`? If a regex, what regex?

Open to ideas here, not sure how far we should go :)

srittau (approved):
Nice, these are very useful additions!

PR #109, opened 2023-10-14
Fix missing `py.typed` files

### LINKED ISSUE #108: partial stubs are missing the partial marker!
Noticed here by @hamdanal https://github.com/typeshed-internal/stub_uploader/issues/93#issuecomment-1732654528
For example, [`types-psycopg2 2.9.21.13`](https://pypi.org/project/types-psycopg2/2.9.21.13/) should include a `py.typed` file with the `partial` keyword in it. But the file is missing.
https://pypi.org/project/types-redis is another example. Looking at `setup.py`, `package_data={'redis-stubs': ['py.typed']}` is found. So the marker is probably physically missing.

Avasam:
Fixes #108

We were accidentally copying `typeshed/stubs/entry-stubs` to `dist/entry-stubs-stubs` instead of `dist/entry-stubs`

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -171,8 +171,11 @@ def copy_stubs(base_dir: str, dst: str) -> None:
             if entry == TESTS_NAMESPACE:
                 # Don't package tests for the stubs
                 continue
-            stub_dir = os.path.join(dst, entry + SUFFIX)
-            shutil.copytree(os.path.join(base_dir, entry), stub_dir)
+            # Don't append the suffix if already present
+            stub_dir = os.path.join(
+                dst, entry if entry.endswith(SUFFIX) else entry + SUFFIX
+            )
+            shutil.copytree(os.path.join(base_dir, entry), stub_dir, dirs_exist_ok=True)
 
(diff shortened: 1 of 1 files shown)


srittau:
This change seems fine, although I'm unsure how this fixes #108?

Avasam:
> This change seems fine, although I'm unsure how this fixes #108?

We were copying the `py.typed` marker file to the wrong location. So it wasn't included in the package.

srittau (approved):

JelleZijlstra:
Does this mean that some of our uploaded stub packages are missing this file? Should we force a new release for those?

AlexWaygood:
Stubs packages generally don't need `py.typed` files. We only started adding them recently to typeshed's packages because we wanted to explicitly mark some of them as `partial`. So it means the new feature has been broken up till now, but hopefully it'll work from this point onwards

Avasam:
> Does this mean that some of our uploaded stub packages are missing this file? Should we force a new release for those?

The 40 stubs with `partial_stub = true` in `METADATA.toml` would be affected. But as @AlexWaygood just mentioned, it wasn't a regression so much as a non-working new feature.

PR #113, opened 2023-11-02
Fix building wheels with py.typed files

srittau:


srittau:
Currently, the integration tests fail, since py.typed is not recognized. I'm not sure, why the tests didn't flag this in #109. Cc @Avasam

srittau:
I think the tests fail locally, because previous runs create py.typed, which subsequent runs then pick up. I'll investigate more.

Avasam:
I think we check for file extensions to only accept `.pyi` and `.md` found in typeshed.
But after running once, a `py.typed` is added, which is then seen on a second run because the run is not cleaned up locally (on CI it doesn't matter since it's a temporary clone of typeshed iirc)

I see two potential solutions:
1. We could allow `py.typed` in our check. But that will allow `py.typed` files to slip in typeshed's source. We may not want that?
2. Add clean-up code.

srittau:
The main problem here is that the integration tests modify files outside the stub uploader directory and don't clean up. I'll open a separate issue.

PR #114, opened 2023-11-02
Switch stub versioning to calver

srittau:
Closes: python/typeshed#10837

### DIFF
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -11,6 +11,7 @@ distribution information.
 """
 
 from __future__ import annotations
+import datetime
 
 from typing import Any
 
@@ -49,8 +50,8 @@ def ensure_specificity(ver: list[int], specificity: int) -> None:
     ver.extend([0] * (specificity - len(ver)))
 
 
-def compute_incremented_version(
-    version_spec: str, published_versions: list[Version]
+def compute_stub_version(
+    version_spec: str, published_versions: list[Version], date: datetime.date
 ) -> Version:
     # The most important postcondition is that the incremented version is greater than
     # all published versions. This ensures that users who don't pin get the most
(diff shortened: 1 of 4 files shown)


srittau:
Integration tests fail, see #113.

stub_uploader/get_version.py
```diff
@@ -72,56 +73,61 @@ def compute_incremented_version(
...
+    ensure_specificity(base_version_parts, stub_specificity)
+    new_version_parts = [*base_version_parts[:-1], int(date.strftime("%Y%m%d"))]
+    new_version = Version(".".join(map(str, new_version_parts)))
+    assert new_version > max_published
+    assert_compatibility(version_base, new_version, is_compatible)
+    return new_version
+
+
+def assert_compatibility(
+    version_base: Version, new_version: Version, is_compatible: bool
```
hauntsaninja, line 119: nit: I like kwargs in cases where is easy to mess up order of args and types all still match

stub_uploader/get_version.py
```diff
@@ -72,56 +73,61 @@ def compute_incremented_version(
...
+        assert new_version not in compatible
 
 
-def determine_incremented_version(metadata: Metadata) -> str:
+def determine_stub_version(metadata: Metadata) -> str:
     published_stub_versions = fetch_pypi_versions(metadata.stub_distribution)
-    version = compute_incremented_version(
-        metadata.version_spec, published_stub_versions
+    version = compute_stub_version(
+        metadata.version_spec, published_stub_versions, datetime.date.today()
```
hauntsaninja, line 131: Should we be careful about timezone? Doesn't matter if this script is always run via Github, but would matter if anyone ever ran by hand

tests/test_unit.py
```diff
@@ -34,45 +35,49 @@ def test_ensure_specificity() -> None:
...
-    assert _incremented_ver("1.1", ["1.1"]) == "1.1.0.1"
-    assert _incremented_ver("1.1", ["1.1.0.4"]) == "1.1.0.5"
-    assert _incremented_ver("1.1", ["1.1.3.4"]) == "1.1.3.5"
-    assert _incremented_ver("1.2.3.4", ["1.2.3.4.5"]) == "1.2.3.4.6"
-    assert _incremented_ver("1.2.3.4.5", ["1.2.3.4.5"]) == "1.2.3.4.5.1"
-
-    # test that we do the max version right
-    assert _incremented_ver("1.2", ["1.1.0.7", "1.2.0.7", "1.3.0.7"]) == "1.3.0.8"
+    assert _stub_ver("1.1", ["1.1"]) == f"1.1.0.{STUB_V}"
+    assert _stub_ver("1.1", ["1.1.0.4"]) == f"1.1.0.{STUB_V}"
```
hauntsaninja, line 77: nit: maybe make some of the published versions look more like dates?

tests/test_unit.py
```diff
@@ -34,45 +35,49 @@ def test_ensure_specificity() -> None:
...
 
     # published equals version spec
-    assert _incremented_ver("1.1", ["1.1"]) == "1.1.0.1"
-    assert _incremented_ver("1.1", ["1.1.0.4"]) == "1.1.0.5"
-    assert _incremented_ver("1.1", ["1.1.3.4"]) == "1.1.3.5"
-    assert _incremented_ver("1.2.3.4", ["1.2.3.4.5"]) == "1.2.3.4.6"
-    assert _incremented_ver("1.2.3.4.5", ["1.2.3.4.5"]) == "1.2.3.4.5.1"
-
-    # test that we do the max version right
-    assert _incremented_ver("1.2", ["1.1.0.7", "1.2.0.7", "1.3.0.7"]) == "1.3.0.8"
```
hauntsaninja, line 75: I think this test is still useful; is only coverage of code path with multiple published versions. RHS just needs to be tweaked

stub_uploader/get_version.py
```diff
@@ -49,8 +50,8 @@ def ensure_specificity(ver: list[int], specificity: int) -> None:
     ver.extend([0] * (specificity - len(ver)))
 
 
-def compute_incremented_version(
-    version_spec: str, published_versions: list[Version]
+def compute_stub_version(
+    version_spec: str, published_versions: list[Version], date: datetime.date
```
hauntsaninja, line 54: ignorable nit: should we try to give a more specific error if given date isn't in the future? (I know we'll error on the `assert new_version > max_published`, but that is a sort of catch-all)

hauntsaninja (commented):
Thanks for working on this!

stub_uploader/get_version.py
srittau: We call it once with the current date, and it would make unit testing harder, so I would prefer not to check this here.

srittau:
Bumpity bump. I'd like to get this in this year if possible.

hauntsaninja (approved):
Thanks for bumping (my github notifications are a mess), this lgtm!

(I still think it would be nice to find a way to give a clearer error if this is run twice in the same day, but that's a non-blocking nit)

PR #117, opened 2023-12-20
Fix integration tests

srittau:
Cf. python/typeshed#11191

### DIFF
--- a/stub_uploader/upload.py
+++ b/stub_uploader/upload.py
@@ -31,8 +31,6 @@ def upload(
 def upload_distribution(
     typeshed_dir: str, distribution: str, commit: str | None, *, dry_run: bool
 ) -> None:
-    already_uploaded = False
-
     print(f"Building stubs for {distribution}... ", end="")
     metadata = read_metadata(typeshed_dir, distribution)
     recursive_verify(metadata, typeshed_dir)
@@ -42,6 +40,8 @@ def upload_distribution(
     except AlreadyUploadedError as e:
         version = str(e.version)
         already_uploaded = True
+    else:
+        already_uploaded = False
 
     if commit:
         update_changelog(
--- a/stub_uploader/upload.py
+++ b/stub_uploader/upload.py
@@ -4,7 +4,7 @@ import os
 import subprocess
 
 from stub_uploader import build_wheel
-from stub_uploader.get_version import determine_stub_version
+from stub_uploader.get_version import AlreadyUploadedError, determine_stub_version
 from stub_uploader.update_changelog import update_changelog
 from stub_uploader.metadata import (
     read_metadata,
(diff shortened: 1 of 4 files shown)


stub_uploader/upload.py
```diff
@@ -25,24 +25,43 @@ def upload(
...
-        if dry_run or not metadata.upload:
-            print(f"skipped")
-        else:
-            subprocess.run(["twine", "upload", os.path.join(temp_dir, "*")], check=True)
-            uploaded_packages.add(distribution)
-            print(f"ok")
-        print()
+    except AlreadyUploadedError as e:
+        version = str(e.version)
+        already_uploaded = True
```
AlexWaygood, line 44: For readability, I would do this rather than setting it to False on line 34. This way it's clearer that the try/except here is the sole determiner of whether this value is True or False 

```suggestion
        already_uploaded = True
    else:
        already uploaded = False
```

tests/test_integration.py
```diff
@@ -50,7 +51,10 @@ def test_build_wheel(distribution: str) -> None:
     "distribution", os.listdir(os.path.join(TYPESHED, THIRD_PARTY_NAMESPACE))
 )
 def test_version_increment(distribution: str) -> None:
-    get_version.determine_stub_version(read_metadata(TYPESHED, distribution))
+    try:
+        get_version.determine_stub_version(read_metadata(TYPESHED, distribution))
+    except AlreadyUploadedError:
+        pass
```
AlexWaygood, line 57: Is it worth asserting that the last part of the version is equal to today's date if it raised AlreadyUploadedError?

AlexWaygood (commented):
Looks good; two small comments:

AlexWaygood (approved):
LGTM

stub_uploader/upload.py
```diff
@@ -25,24 +25,43 @@ def upload(
...
+        )
+
+    temp_dir = build_wheel.main(typeshed_dir, distribution, version)
+    print(f"ok, version {version}")
+
+    print(f"Uploading stubs for {distribution}... ", end="")
+    if dry_run or not metadata.upload:
+        print("skipped")
+    elif already_uploaded:
+        print("already uploaded, skipped")
```
hauntsaninja, line 62: I'm a little worried about this skip. This could result in us dropping updates to specific stub packages / the state being confusing, because it won't prevent update of `data/last_typeshed_commit.sha1`

srittau: Good point. It shouldn't be a problem in practice, since the upload only runs once per day, but I'll investigate a bit more tomorrow if we can make this safer. Alternatively we could switch back to the plan of just advancing to the next free date. I.e. use 20231216 if 20231214 and 20231215 were already uploaded, even if it's still the 14th.

PR #118, opened 2023-12-21
Use future dates if necessary

srittau:
This changes the logic so that multiple package versions
can be uploaded per day. In that case, the next date is
used. This should only happen in practice if an upload
was triggered manually.

### DIFF
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -30,12 +30,6 @@ RETRY_ON = [429, 500, 502, 503, 504]
 TIMEOUT = 3
 
 
-class AlreadyUploadedError(Exception):
-    def __init__(self, version: Version) -> None:
-        super().__init__(f"Version {version} was already uploaded")
-        self.version = version
-
-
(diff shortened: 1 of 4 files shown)


srittau:
Cf. #117, @hauntsaninja

stub_uploader/get_version.py
```diff
@@ -102,31 +96,48 @@ def compute_stub_version(
...
+    # If the base version changed, we should always be able to use the current date.
+    if base_version_changed:
+        return date
+
+    # Make sure that the last published release is not unreasonable large.
+    assert int(date.strftime("%Y%m%d")) + 14 > max_published.release[-1]
+
+    while int(date.strftime("%Y%m%d")) <= max_published.release[-1]:
+        date = date + datetime.timedelta(days=1)
+    return date
```
srittau, line 138: An (easier) alternative here would be to just count up, e.g. 20231031 -> 20231032, but I think only using real dates is a bit less confusing.

hauntsaninja (approved):
This looks good to me! Thanks again!

PR #119, opened 2024-01-08
Encapsulate PackageData; use pathlib

srittau:
This is a preliminary refactoring for fixing python/typeshed#11254. The idea is that the new class `PackageData` encapsulates all data concerning packages and their contents. This allows us later to find the top-level non-namespace packages, instead of just the top-level packages as we are doing now.

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -245,7 +245,7 @@ def collect_package_data(base_path: Path) -> PackageData:
                             f"Only stub files are allowed, not {entry.name!r}"
                         )
                 continue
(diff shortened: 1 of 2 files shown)


srittau:
@Avasam Could you review this? It should be fairly self-explanatory, even if you are not well-versed with the stub uploader.

Avasam:
I can do a proper review tomorow. But a first overview from my phone this looks good. This is an area of stub_uploader I'm a bit familiar with.

Avasam (approved):
Pretty much a 1:1 mapping of existing logic and looks cleaner. Looking forward to follow-up PRs!

JelleZijlstra (approved):

Issue #112, opened 2023-10-23
setuptools v69 includes type information by default


Avasam:
I'd like to bring some attention to https://github.com/pypa/setuptools/pull/4021 . Since [setuptools v69](https://setuptools.pypa.io/en/stable/history.html#v69-0-0), `.pyi` and `py.typed` should be picked up by default. It is experimental, but I think it's worth looking into, as it might simplify `stub_uploader`'s logic in some places.

Avasam:
Thanks for the quick revert.
I tried a few cases but I guess I missed that.
Let's make sure to have tests to cover that case for the next attempt ^^

srittau:
> Let's make sure to have tests to cover that case for the next attempt ^^

It's on my to do list for today or tomorrow, unless you're working on something?

Avasam:
> > Let's make sure to have tests to cover that case for the next attempt ^^
> 
> It's on my to do list for today or tomorrow, unless you're working on something?

Go ahead. I got my hands full with my dayjob

PR #120, opened 2024-01-08
Refactor collect_package_data to make logic clearer

srittau:
Depends on #119


stub_uploader/build_wheel.py
```diff
@@ -228,34 +228,40 @@ def collect_package_data(base_path: Path) -> PackageData:
...
+            pkg_name = entry.name
+            stub_files = find_stub_files(str(entry))
         else:
-            if entry.name == TESTS_NAMESPACE:
-                continue
-            pkg_name = entry.name + SUFFIX
-            package_data[pkg_name] = find_stub_files(str(entry))
-        package_data[pkg_name].append(META)
+            raise ValueError(f"Only stub files are allowed, not {entry.name!r}")
+        package_data[pkg_name + SUFFIX] = [*stub_files, META]
```
Avasam, line 242: Maybe you won't wanna do that now, but since you're touching this area of code I'll mention it:
With https://github.com/typeshed-internal/stub_uploader/issues/112 (setuptools v69) , the `stub_files` logic could become unnecessary.

According to https://setuptools.pypa.io/en/stable/userguide/miscellaneous.html#controlling-files-in-the-distribution this is an opt-out feature, so you shouldn't need to add anything new either.

srittau: Let's punt this to a separate PR to keep this as a simple refactoring.

Avasam (approved):
New logic checks out and is easier to reason about.

Issue #121, opened 2024-01-12
Fix py.typed file creation


srittau:
Currently, when building a distribution, if a py.typed file needs to be created, this happens during the `generate_setup_file()` step. The file is added to the typeshed directory and will get copied to the tmp dir during the `copy_stubs()` step. This is problematic for a few reasons:

* Conceptually, `generate_setup_file()` makes no modifications. It just returns the setup.py file as a string. The exception being creating the `py.typed` file if necessary.
* The build process now modifies the typeshed directory, when it should be considered "read-only".
* This also means that the build process can't be started twice before cleaning the typeshed dir.
* This is especially problematic in local development environments, where your local typeshed directory gets modified and you can't run the integration tests twice.

The solution is to create a new step between `copy_stubs()` and `copy_changelog()` that creates the `py.typed` marker directly in the tmpdir. I'll send a PR after the current batch of PRs was processed.

Avasam:
Btw I found why `py.typed` files are missing from namespace packages. The path added to `package_data` is just wrong. Something like this outta do it:
![image](https://github.com/typeshed-internal/stub_uploader/assets/1350584/3c10dbff-9381-408e-80f3-1536525af502)

PR #122, opened 2024-01-12
Fix py.typed in namespace packages

srittau:
Previously, the py.typed file was incorrectly put in the namespace package, instead of the non-namespace sub-packages.


srittau:
Cc @Avasam

srittau:
Could someone review this? The google packages are still broken.

tests/test_integration.py
```diff
@@ -169,3 +164,35 @@ def test_verify_requires_python() -> None:
...
+        ("Pillow", ["PIL-stubs"]),
+        ("protobuf", ["google-stubs.protobuf"]),
+        ("google-cloud-ndb", ["google-stubs.cloud.ndb"]),
+    ],
+)
+def test_pkg_data_non_namespace_packages(
+    distribution: str, expected_packages: list[str]
+) -> None:
+    pkg_data = build_wheel.collect_package_data(THIRD_PARTY_PATH / distribution)
+    assert pkg_data.top_level_non_namespace_packages == expected_packages
```
Avasam, line 198: I just noticed, I think there's no test to validate that `py.typed` is added to the right location(s). Could we add unit tests for the `if metadata.partial` branch of `generate_setup_file` calling `add_partial_markers`?

Avasam (changes requested):
I have one optional suggestion and one request for an additional test concerning `py.typed` files. Otherwise LGTM.

tests/test_integration.py
srittau: Good idea, but I think it makes more sense as part of #121, considering the current command/query mixture.

Avasam (approved):

srittau:
Unfortunately, it seems that py.typed is now dropped completely from the package. I'll look into it later.

srittau:
It's been a few weeks, but I also remember testing this locally where it seemed to work. Probably best to just work on #121 instead of trying to find this bug.

Avasam:
@srittau Should I wait before merging PRs relating to partial stubs? (ie https://github.com/python/typeshed/pull/11255/files ) 
iirc would affect type-checking users that also use modules missing from typeshed.

srittau:
I would actually be interested if it also affects partial non-namespace packages.

Avasam:
> I would actually be interested if it also affects partial non-namespace packages.

pyopenssl was released 20h ago and contains the marker file
https://pypi.org/project/types-pyOpenSSL/#history
image

I believe this PR was merged 29h ago?

If that's still too close/unsure, you'll know for sure once tensorflow (modified 3h ago) deploys.

---

I think I understand what you're thinking of: maybe only namespace packages are missing the `py.typed` file. Which could be caused, for example, by an incorrect path added to `package_data[package]`

PR #124, opened 2024-01-29
Update and pin black

srittau:


AlexWaygood (approved):
LGTM. We can probably update the `requests` pin, but doesn't need to be done in this PR.

PR #126, opened 2024-01-31
Use pypa/build

### LINKED ISSUE #125: Calling `setup.py` directly is deprecated
See the following warning when called:
```
C:\Program Files\Python39\lib\site-packages\setuptools\_distutils\cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
!!

        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
        ********************************************************************************

!!
```

Avasam:
From a local test this *looks correct*, but I may have missed something. Attempt to close #125 

Doc: https://build.pypa.io/en/latest/

### DIFF
--- a/.github/workflows/test_api_token.yml
+++ b/.github/workflows/test_api_token.yml
@@ -22,12 +22,12 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine
+        pip install build setuptools wheel twine
     - name: Build and publish
       env:
         TWINE_USERNAME: __token__
         TWINE_PASSWORD: ${{ secrets.TYPESHED_BOT_API_TOKEN }}
       run: |
         cd data/empty_package
-        python setup.py bdist_wheel ${{ github.event.inputs.increment }}
+        python -m build --wheel ${{ github.event.inputs.increment }}
         twine upload dist/*
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,18 @@
+# scripts
 packaging
 requests < 2.30.0
 tomli
 
+# build and upload
+build
 setuptools
-wheel
 twine
+wheel
 
+# tests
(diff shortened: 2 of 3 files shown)


.github/workflows/test_api_token.yml
```diff
@@ -22,12 +22,12 @@ jobs:
...
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine
+        pip install build setuptools wheel twine
     - name: Build and publish
       env:
         TWINE_USERNAME: __token__
         TWINE_PASSWORD: ${{ secrets.TYPESHED_BOT_API_TOKEN }}
       run: |
         cd data/empty_package
-        python setup.py bdist_wheel ${{ github.event.inputs.increment }}
```
Avasam, line 32: What was this meant to do?
```
PS C:\Users\Avasam\AppData\Local\Temp\tmpk_pfo33j> python setup.py bdist_wheel 0
invalid command name '0'
```

hauntsaninja: You can't upload wheels with the same version: https://github.com/typeshed-internal/stub_uploader/blob/main/data/empty_package/setup.py#L5

Avasam: Ah! I see how it's used now. Ok, I'll have to account for that. Setting as draft for now.

Avasam:
I added the `--no-isolation` flag to speedup the build and avoid having to create a venv everytime. I believe this running in the CI should be fine? If you believe there's an issue with that, security or otherwise, the flag can be removed.

srittau (approved):

srittau:
This seems to work fine: https://pypi.org/project/types-networkx/#files

PR #129, opened 2024-03-17
Fix creating py.typed files

srittau:
* Create the py.typed files in the build directory instead of the typeshed directory. Closes: #121
* Fix creating py.typed files in namespace sub-packages.
* Always create py.typed files, not only partial ones.


srittau:
Cc @Avasam

stub_uploader/build_wheel.py
```diff
@@ -404,21 +416,29 @@ def main(
     created after uploading it.
     """
     ts_data = read_typeshed_data(Path(typeshed_dir))
+    # TODO: Merge BuildData and PackageData and encapsulate MetaData.
+    # See also https://github.com/typeshed-internal/stub_uploader/issues/123.
```
Avasam, line 420: The doc messes up with some url interpreters

```suggestion
    # See also https://github.com/typeshed-internal/stub_uploader/issues/123 .
```
or
```suggestion
    # See also https://github.com/typeshed-internal/stub_uploader/issues/123
```

---

For example:
image

Avasam (approved):
This looks correct!
image

PR #130, opened 2024-03-25
Move commit reading into typeshed_data

srittau:
Use Path object for tmpdir

Use subprocess.run(cwd) instead of changing the cwd manually


srittau:
Pingity ping. This is a prerequisite for further refactorings to make the stub uploader more robust to change.

JelleZijlstra (approved):
Looks good. I like that we're no longer using `os.chdir`; it's cleaner to use absolute paths.

PR #134, opened 2024-04-14
Run tests using Python 3.12 in CI

AlexWaygood:
And fix a tarfile DeprecationWarning that's new in Python 3.12, and has been showing up when the `stub_uploader` tests are run in typeshed's CI.

The tarfile extraction filters were backported to all security branches as part of PEP 706. Nonetheless, I used the method the docs advise to ensure backwards compatibility with versions of Python that don't have these extraction filters (e.g. if you're running with Python 3.11.4, the feature exists, but if you're running with Python 3.11.3, it doesn't). See https://docs.python.org/3.12/library/tarfile.html#supporting-older-python-versions.

### DIFF
--- a/.github/workflows/check_scripts.yml
+++ b/.github/workflows/check_scripts.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: '3.12'
       - uses: actions/checkout@v4
       - run: pip install -r requirements.txt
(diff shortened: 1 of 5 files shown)


AlexWaygood:
I didn't bump Python to 3.12 in the workflows that actually push stubs to PyPI. That felt slightly riskier than just bumping the version used to run the tests? Or at least, it felt like it deserved its own PR.

srittau (approved):

PR #135, opened 2024-04-15
Add django-stubs to allowlist

srittau:
Cf. python/typeshed#11709


Viicos:
Yeah turns out we are facing the same issue as https://github.com/python/typeshed/pull/10919 :/

srittau:
This seems unnecessary.

Viicos:
Thanks, reading from https://github.com/python/typeshed/pull/10919 seems like I can also update my pending `typeshed` PR

PR #138, opened 2024-06-08
Add `tree-sitter` to the list of allowed external packages

sobolevn:
Refs https://github.com/python/typeshed/pull/12118

### DIFF
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -222,6 +222,7 @@ EXTERNAL_REQ_ALLOWLIST = {
     "referencing",
     "torch",
     "urllib3",
+    "tree-sitter",
 }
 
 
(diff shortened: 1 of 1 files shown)


srittau (approved):

AlexWaygood (approved):

Akuli:
I don't like how we need to add relatively obscure packages to a global allowlist, but I'm not sure if there is a better solution. I guess we just need to carefully avoid step 6 of https://github.com/typeshed-internal/stub_uploader/pull/61#discussion_r979327370.

That said, it would be nice to document the security aspects of stub_uploader somewhere, maybe to a markdown file in this repo, instead of referring to old PR comments whenever security comes up. I might give it a try within the next few weeks. 

Security is IMO the most important thing for stub_uploader to get right, because a malicious `types_requests` could very quickly gain access to many dev machines, and from there to many production servers and such.

PR #139, opened 2024-07-10
Use a link in outro comment

sobolevn:
This is how it currently looks like:
Снимок экрана 2024-07-10 в 13 54 13

When I wanted to click on that commit I realized that I have to:
- Go to `typeshed`
- Select some other commit page
- Copy-paste the needed commit from the outro

It would be better with a link :)


AlexWaygood (approved):
Thanks! I've wanted this in the past too.

srittau (approved):

Issue #90, opened 2023-02-16
Allowing more external dependencies for stubs


hauntsaninja:
Currently, we're quite strict about what external dependencies stubs can use. We have a small allowlist here:
https://github.com/typeshed-internal/stub_uploader/blob/b81ba3c5214667f1dfb1decb0c97b83214553833/stub_uploader/metadata.py#L166

Briefly, the reason for this is a) Python packages can execute arbitrary code at install time, b) type checkers sometimes install stub packages automatically, c) stub packages are quite popular, d) users likely expect stub packages to be inert

So what would it take to remove the allowlist? We already have an important additional check in stub uploader: we ensure that external dependencies of a stub package must be a dependency of the upstream package. However, there's still a hole, in that stub dependencies of stub packages are not currently checked.

To spell things out, this is the scenario that we're concerned about:

- Package foo exists, that depends on bar
- typeshed adds types-foo, add dependency on bar
- In order to make types-foo useful, we need bar, so we do some security review of bar / the bar maintainer and add it to the allowlist
- bar (gets hacked and) adds a dependency on evil
- types-foo users are affected, while this isn't good we're okay with it, since those are assumed to be a subset of foo users
- Someone sneaks past typeshed maintainers and adds a dependency on types-foo to types-requests (wouldn't be too hard to do)
- types-requests users are affected, this is really bad

Once we plug this hole, we could maybe get rid of the allowlist or have much more relaxed criteria.

This has been discussed in a couple places, mainly https://github.com/typeshed-internal/stub_uploader/pull/61#discussion_r979316520. I'm writing this up here as a way to easily communicate the current status quo to typeshed contributors.

Plugging the hole is a pretty easy change to make to stub_uploader, see diff here:
<details>

```
diff --git a/stub_uploader/metadata.py b/stub_uploader/metadata.py
index 85d0643..56c5448 100644
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -61,7 +61,7 @@ class Metadata:
     def requires_typeshed(self) -> list[Requirement]:
         reqs = self._unvalidated_requires_typeshed
         for req in reqs:
-            verify_typeshed_req(req)
+            verify_typeshed_req(req, self.upstream_distribution)
         return reqs
 
     @property
@@ -145,7 +145,7 @@ def strip_types_prefix(dependency: str) -> str:
     return dependency.removeprefix(TYPES_PREFIX)
 
 
[...]

PR #140, opened 2024-07-15
Add documentation of security implications

srittau:
The maintainers section is still missing, but the remaining sections are up for scrutiny.

Cc @AlexWaygood @hauntsaninja @Avasam

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -45,8 +45,8 @@ allowed:
 
 Stub packages are uploaded to PyPI by the PyPI user
 [typeshed_bot](https://pypi.org/user/typeshed_bot/) using an API token.
-The packages are owned solely by that user, so that no other user can
-upload new versions of the packages, in case the account of a
+The packages are owned solely by that user, so that no other PyPI user can
+upload new versions of the packages, in case the PyPI account of a
 maintainer becomes compromised.
 
 ## Reporting a Vulnerability
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,7 +14,7 @@ is also the reason that this repository is part of a separate GitHub
 organization and not the [python](https://github.com/python) or
 [psf](https://github.com/psf) organizations, like typeshed is.
 
-Typeshed maintainers without commit right are still considered maintainers
+Typeshed maintainers without commit rights are still considered maintainers
 of the stub uploader and are encouraged to manage issues and send and review
 pull requests.
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -32,3 +32,10 @@ allowed:
   upstream package, this does not introduce an additional security liability.
 * Dependencies that are explicitly allowlisted in the stub uploader. These
   dependencies are vetted to be from a trusted source.
+
+## PyPI Uploads
+
+Stub packages are uploaded to PyPI by the PyPI user
+[typeshed_bot](https://pypi.org/user/typeshed_bot/) using an API token.
+The packages are owned solely by that user, so that no other user can
+upload new versions of the packages, in case an account gets compromised.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -29,7 +29,7 @@ modules must be imported.
 
 ## Dependencies
 
-Another possible attack vector are dependencies of stub packages.
+Another possible attack vector is dependencies of stub packages.
 A compromised dependency can have a similar effect to when a stub package
 gets compromised directly. Therefore, only certain dependencies are
 allowed:
@@ -46,7 +46,8 @@ allowed:
 Stub packages are uploaded to PyPI by the PyPI user
 [typeshed_bot](https://pypi.org/user/typeshed_bot/) using an API token.
 The packages are owned solely by that user, so that no other user can
-upload new versions of the packages, in case an account gets compromised.
+upload new versions of the packages, in case the account of a
+maintainer becomes compromised.
 
 ## Reporting a Vulnerability
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -8,7 +8,15 @@ stub packages, this could have significant security implications.
 
 ## Maintainers
 
-TBD
+Only a small number of typeshed maintainers have commit rights for this
+repository to minimize the attack vector in case of hacked accounts. This
+is also the reason that this repository is part of a separate GitHub
+organization and not the [python](https://github.com/python) or
+[psf](https://github.com/psf) organizations, like typeshed is.
+
+Typeshed maintainers without commit right are still considered maintainers
+of the stub uploader and are encouraged to manage issues and send and review
+pull requests.
 
 ## Typeshed Data
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -12,7 +12,7 @@ TBD
 
 ## Typeshed Data
 
-To ensure that a compromised typeshed repository can't lead to copromised
+To ensure that a compromised typeshed repository can't lead to compromised
 stub packages, all typeshed data is verified by the stub uploader before
 building packages. The stub uploader ensures that only stub and metadata
 files are added to the stub packages. This also means that no code from the
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -47,3 +47,7 @@ Stub packages are uploaded to PyPI by the PyPI user
 [typeshed_bot](https://pypi.org/user/typeshed_bot/) using an API token.
 The packages are owned solely by that user, so that no other user can
 upload new versions of the packages, in case an account gets compromised.
+
+## Reporting a Vulnerability
+
+To report an issue, go to https://github.com/typeshed-internal/stub_uploader/security. We commit to respond to any issue within 14 days and promptly fix any problems.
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -0,0 +1,34 @@
...
+Considering the high trust, quick turnaround, and automated installation of
+stub packages, this could have significant security implications.
+
+## Maintainers
+
+TBD
+
+## Typeshed Data
+
+To ensure that a compromised typeshed repository can't lead to copromised
```
JelleZijlstra, line 15:
```suggestion
To ensure that a compromised typeshed repository can't lead to compromised
```

JelleZijlstra (commented):
Thanks, great idea to add this!

Another area to cover could be the security of the upload process itself: what are we doing to prevent an issue where a malicious party could gain the ability to upload stub packages using the uploader's identity? (Relevant real-world example: https://blog.pypi.org/posts/2024-07-08-incident-report-leaked-admin-personal-access-token/.)

srittau:
Ironically, I forgot to ping @Akuli, although it was #138 that prompted me to write this. Sorry!

@JelleZijlstra What can we do about that? In a related matter: I can't check, but do we require 2FA to be enabled for this org/repository? If not, we should definitely enable it.

ilevkivskyi:
> but do we require 2FA to be enabled for this org/repository?

I am 90% sure I enabled it a while ago, I will double-check now.

JelleZijlstra:
> What can we do about that?

Haven't thought too much about it, but it looks like stub packages are owned only by typeshed-bot (e.g., https://pypi.org/project/types-requests/). That's good because it means that if someone gains access to my PyPI account, they can't use it to upload a malicious stubs package. However, we should also review what kind of authentication typeshed-bot uses to upload packages.

ilevkivskyi:
> I am 90% sure I enabled it a while ago, I will double-check now.

Actually it turns out "Require two-factor authentication for everyone in the typeshed-internal organization" was off. I just turned it on. (I also checked that all four members have 2FA enabled ;-))

ilevkivskyi:
> what kind of authentication typeshed-bot uses to upload packages

It uses a PyPI token. FWIW we can rotate it from time to time. (There is even a GHA to test that PyPI token works.)

srittau:
We should probably document the typeshed-bot user in this file as well. Commits to the PR or text suggestions in the comments welcome. Otherwise I will look into this tomorrow.

SECURITY.md
```diff
@@ -0,0 +1,34 @@
...
+A compromised dependency can have a similar effect to when a stub package
+gets compromised directly. Therefore, only certain dependencies are
+allowed:
+
+* Dependencies on other stub packages created by typeshed.
+* Dependencies on packages the upstream package depends on – even recursively.
+  Since it's likely that a stub package gets installed alongside the
+  upstream package, this does not introduce an additional security liability.
+* Dependencies that are explicitly allowlisted in the stub uploader. These
+  dependencies are vetted to be from a trusted source.
```
Akuli, line 34: It would be nice to expand upon this and actually describe and explain the possible attack in https://github.com/typeshed-internal/stub_uploader/pull/61#discussion_r979327370. This kind of hints at how the attack goes, but IMO does not describe it in enough detail. It is the biggest security issue in stub_uploader I'm aware of, and typeshed maintainers need to be aware of it.

Avasam:
I got nothing to add. You covered everything I was already aware of 👍

SECURITY.md
srittau: I'd like to hold off documenting this here at the moment. I kept this section quite vague for the moment, mostly because I think we should resolve and implement #90 first and can then discuss the measures and reasons in this document.

srittau:
I've added the maintainers section, this is ready for a full review now. As mentioned above, I'd like to punt @Akuli's suggestion to describe the attack to later when we've decided what to do in the future in #90.

SECURITY.md
```diff
@@ -0,0 +1,49 @@
+# Security Implication of the Typeshed Stub Uploader
```
JelleZijlstra, line 1: A SECURITY.md should by convention include a way to privately report security vulnerabilities. For typing-extensions (https://github.com/python/typing_extensions/blob/main/SECURITY.md), I redirected reports to GitHub's private reporting feature, but for this repo I don't have permissions to turn on that feature.

So I'd suggest:

- Someone with the right permissions (@ilevkivskyi?) should go to https://github.com/typeshed-internal/stub_uploader/security and turn on "Private vulnerability reporting"
- In this PR we should add a line saying that any vulnerabilities should be reported to https://github.com/typeshed-internal/stub_uploader/security

We should also add a SECURITY.md to the typeshed repo pointing here.

SECURITY.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Maintainers
+
+Only a small number of typeshed maintainers have commit rights for this
+repository to minimize the attack vector in case of hacked accounts. This
+is also the reason that this repository is part of a separate GitHub
+organization and not the [python](https://github.com/python) or
+[psf](https://github.com/psf) organizations, like typeshed is.
+
+Typeshed maintainers without commit right are still considered maintainers
```
JelleZijlstra, line 17:
```suggestion
Typeshed maintainers without commit rights are still considered maintainers
```

SECURITY.md
ilevkivskyi: OK, I can do this later today, or @JukkaL should be able to do this.

srittau: I've added such a section to the document for now, mostly taking the wording from typing-extension.

ilevkivskyi: OK, I enabled "Private vulnerability reporting" for this repo.

srittau:
Are there any more notes?

JelleZijlstra (approved):

SECURITY.md
```diff
@@ -0,0 +1,53 @@
...
+To ensure that a compromised typeshed repository can't lead to compromised
+stub packages, all typeshed data is verified by the stub uploader before
+building packages. The stub uploader ensures that only stub and metadata
+files are added to the stub packages. This also means that no code from the
+typeshed repository must be executed while building packages, and no
+modules must be imported.
+
+## Dependencies
+
+Another possible attack vector are dependencies of stub packages.
```
AlexWaygood, line 32: "Possible attack vector" is a singular noun

```suggestion
Another possible attack vector is dependencies of stub packages.
```

SECURITY.md
```diff
@@ -0,0 +1,53 @@
...
+  upstream package, this does not introduce an additional security liability.
+* Dependencies that are explicitly allowlisted in the stub uploader. These
+  dependencies are vetted to be from a trusted source.
+
+## PyPI Uploads
+
+Stub packages are uploaded to PyPI by the PyPI user
+[typeshed_bot](https://pypi.org/user/typeshed_bot/) using an API token.
+The packages are owned solely by that user, so that no other user can
+upload new versions of the packages, in case an account gets compromised.
```
AlexWaygood, line 49:
```suggestion
upload new versions of the packages, in case the account of a
maintainer becomes compromised.
```

AlexWaygood (approved):

SECURITY.md
```diff
@@ -0,0 +1,54 @@
...
+* Dependencies that are explicitly allowlisted in the stub uploader. These
+  dependencies are vetted to be from a trusted source.
+
+## PyPI Uploads
+
+Stub packages are uploaded to PyPI by the PyPI user
+[typeshed_bot](https://pypi.org/user/typeshed_bot/) using an API token.
+The packages are owned solely by that user, so that no other user can
+upload new versions of the packages, in case the account of a
+maintainer becomes compromised.
```
Akuli, line 50: Couldn't a maintainer account simply edit CI config to store the token somewhere? Or is this about maintainers who do not have commit access, like me?

SECURITY.md
Akuli: Maybe:

```suggestion
  dependencies are vetted to be from a trusted source.

Having one global allowlist is not ideal, because it allows somewhat obscure
dependencies to be added to popular stub packages whose security matters
the most. This may be changed in the near future.
```

SECURITY.md
JelleZijlstra: I think it's about the PyPI accounts, not GitHub accounts. If a GitHub account with commit access (e.g., mine) gets compromised, it could obviously be used to upload a malicious package just by changing the code. But if someone compromises my PyPI account, they can't use it to upload malicious stub packages. (They could use it to upload a malicious version of typing-extensions, which is probably worse, but that's a different repo than this one.)

SECURITY.md
```diff
@@ -0,0 +1,54 @@
...
+  upstream package, this does not introduce an additional security liability.
+* Dependencies that are explicitly allowlisted in the stub uploader. These
+  dependencies are vetted to be from a trusted source.
+
+## PyPI Uploads
+
+Stub packages are uploaded to PyPI by the PyPI user
+[typeshed_bot](https://pypi.org/user/typeshed_bot/) using an API token.
+The packages are owned solely by that user, so that no other user can
+upload new versions of the packages, in case the account of a
```
Akuli, line 49:
```suggestion
The packages are owned solely by that user, so that no other PyPI user can
upload new versions of the packages, in case the PyPI account of a
```

Akuli (approved):

PR #143, opened 2024-08-03
Use SPDX license string for stubs packages

cdce8p:
The SPDX identifier for the [Apache License 2.0](https://spdx.org/licenses/Apache-2.0.html) is `Apache-2.0`. [PEP-639](https://peps.python.org/pep-0639/) is close to being finalized and it will standardize SPDX license expressions for packages. It's possible though to start using them today.

It might make sense to force update all stub packages in this case.

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -73,7 +73,7 @@ setup(name=name,
       install_requires={requires},
       packages={packages},
       package_data={package_data},
-      license="Apache-2.0 license",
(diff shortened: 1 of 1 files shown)


srittau (approved):

srittau:
Thanks!

JelleZijlstra:
> It might make sense to force update all stub packages in this case.

I think we shouldn't do that before the PEP is accepted. After that happens, we can see how many stub packages haven't been changed yet.

PR #144, opened 2024-09-26
Validate the versions in pyproject.toml

srittau:
Use packaging.Version instead of a regexp to validate these are valid Python versions.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -155,9 +155,6 @@ def test_recursive_verify(distribution: str) -> None:
 
 def test_read_typeshed_data() -> None:
     ts_data = read_typeshed_data(Path(TYPESHED))
-    assert re.match(r"^\d+\.\d+\.\d+$", ts_data.mypy_version)
-    assert re.match(r"^\d+\.\d+\.\d+$", ts_data.pyright_version)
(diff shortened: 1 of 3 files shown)


srittau:
This reduces the attack vector from a maliciously formed version string in typeshed's pyproject file. It also fixes python/typeshed#12692, where the regexp we used was overly strict.

srittau:
Cc @sobolevn

sobolevn (approved):
Thank you!

tests/test_integration.py
```diff
@@ -155,9 +155,6 @@ def test_recursive_verify(distribution: str) -> None:
 
 def test_read_typeshed_data() -> None:
     ts_data = read_typeshed_data(Path(TYPESHED))
-    assert re.match(r"^\d+\.\d+\.\d+$", ts_data.mypy_version)
```
JelleZijlstra, line 158: Is it worth adding a test that we fail if one of these version strings is invalid?

srittau: @JelleZijlstra Done. I've now reduced the integration test to just verify that the `read_typeshed_data` function returns and added unit tests for it instead.

JelleZijlstra (approved):

PR #145, opened 2024-10-02
Add backports.zoneinfo to allowed dependencies

srittau:
Also improve error message when dependency is missing

backports.zoneinfo is maintained by CPython core developer @pganssle, so it's uncritical. Required for python/typeshed#12706.

### DIFF
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -214,6 +214,7 @@ EXTERNAL_REQ_ALLOWLIST = {
     "Pillow",
     "Werkzeug",
     "arrow",
+    "backports.zoneinfo",
     "click",
     "cryptography",
     "django-stubs",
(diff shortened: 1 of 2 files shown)


stub_uploader/metadata.py
```diff
@@ -289,7 +290,7 @@ def verify_external_req(
 
     if req.name not in EXTERNAL_REQ_ALLOWLIST and not _unsafe_ignore_allowlist:
         raise InvalidRequires(
-            f"Expected dependency {req} to be present in the allowlist"
+            f"Expected dependency {req.name} to be present in the allowlist"
```
AlexWaygood, line 293: I think it's often unclear to our contributors which allowlist this is referring to. This maybe isn't much better, but it's a small improvement:
```suggestion
            f"Expected dependency {req.name} to be present in the stub_uploader allowlist"
```

AlexWaygood (approved):
Seems like my suggestion broke a few tests that now need to be tweaked, but LGTM other than that

ilevkivskyi (approved):

PR #149, opened 2024-10-19
Fix version date calculation in corner cases

srittau:
Previously, if the latest published stub version was higher
than the upstream version (which could happen with our old
versioning scheme), we always used the current date as
last component. This failed when uploading the same package
twice in a day – but also in our stub_uploader tests, which
simulate a new upload.

### DIFF
--- a/stub_uploader/get_version.py
+++ b/stub_uploader/get_version.py
@@ -50,8 +50,9 @@ def ensure_specificity(ver: list[int], specificity: int) -> None:
     ver.extend([0] * (specificity - len(ver)))
 
 
+# `today` can be passed as an argument for testing purposes.
 def compute_stub_version(
-    version_spec: Specifier, published_versions: list[Version], date: datetime.date
+    version_spec: Specifier, published_versions: list[Version], today: datetime.date
 ) -> Version:
     # The most important postcondition is that the incremented version is greater than
     # all published versions. This ensures that users who don't pin get the most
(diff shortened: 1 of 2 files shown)


srittau:
Best reviewed commit by commit.

stub_uploader/get_version.py
```diff
@@ -123,19 +124,25 @@ def compute_stub_version(
...
+    last_date = max_published.release[-1]
+
     # If the base version changed, we should always be able to use the current date.
     if base_version_changed:
-        return date
+        return today
 
     # Make sure that the last published release is not unreasonable large.
-    assert int(date.strftime("%Y%m%d")) + 14 > max_published.release[-1]
+    assert int(today.strftime("%Y%m%d")) + 14 > last_date
```
JelleZijlstra, line 141: Where does 14 come from here? Is it just arbitrary?

srittau: Yes. It's very unlikely we publish more than 14 versions in a day.

JelleZijlstra: While 13 versions are just a little unlikely, I guess :)

JelleZijlstra (approved):

stub_uploader/get_version.py
AlexWaygood: > While 13 versions are just a little unlikely, I guess :)

a little unlikely, but very unlucky

PR #151, opened 2024-11-10
Add pyproj to external dependencies allow list

hamdanal:
This is an important dependency of geopandas which I am adding to typeshed in https://github.com/python/typeshed/pull/12990. It is in the top 1000 PyPI packages https://hugovk.github.io/top-pypi-packages/

### DIFF
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -231,6 +231,7 @@ EXTERNAL_REQ_ALLOWLIST = {
     "torch",
     "tree-sitter",
     "urllib3",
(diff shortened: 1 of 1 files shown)


stub_uploader/metadata.py
```diff
@@ -231,6 +231,7 @@ def verify_typeshed_req(req: Requirement) -> None:
     "torch",
     "tree-sitter",
     "urllib3",
+    "pyproj",
```
srittau, line 234: Please note that the list is sorted alphabetically (except that capitalized names are first).

PR #152, opened 2024-11-11
Check for runtime requirements when depending on external stub packages

hamdanal:
When a typeshed package depends on an external stub-only package, its runtime package must depend on the corresponding runtime package of the stub-only dependency. For example, [types-geopandas](https://github.com/python/typeshed/pull/12990) requiring `pandas-stubs` means [geopandas](https://github.com/geopandas/geopandas/blob/22345152d9d99a47c88808f763e27bf832ef81e4/pyproject.toml#L27) must require pandas, not pandas-stubs.

Unblocks: https://github.com/python/typeshed/pull/12990

### DIFF
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -234,6 +234,13 @@ EXTERNAL_REQ_ALLOWLIST = {
     "urllib3",
 }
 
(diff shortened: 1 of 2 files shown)


srittau (approved):
Thanks!

PR #159, opened 2025-02-14
Allow types-foo to depend on foo

Avasam:
This is relevant `setuptools` in https://github.com/python/typeshed/pull/13369#discussion_r1903394374 and might be relevant to `pywin32` later as well.

Relates to #90


hauntsaninja:
I don't think type checkers will resolve this correctly.

stub_uploader/metadata.py
```diff
@@ -350,9 +350,11 @@ def verify_external_req_stubs_require_its_runtime(
...
     runtime_req_name = EXTERNAL_RUNTIME_REQ_MAP.get(req.name, req.name)
 
-    if not runtime_in_upstream_requires(
-        req, data
-    ) and not runtime_in_upstream_sdist_requires(req, data):
+    if (
+        runtime_req_name != upstream_distribution  # Allow `types-foo` to require `foo`
+        and not runtime_in_upstream_requires(req, data)
+        and not runtime_in_upstream_sdist_requires(req, data)
+    ):
```
Avasam, line 357:
```suggestion
    ):
```

Avasam, line 357: Alternative syntax:
```suggestion
    if not (
        runtime_req_name == upstream_distribution  # Allow `types-foo` to require `foo`
        or runtime_in_upstream_requires(req, data)
        or runtime_in_upstream_sdist_requires(req, data)
    ):
```

stub_uploader/metadata.py
```diff
@@ -350,9 +350,11 @@ def verify_external_req_stubs_require_its_runtime(
 
     runtime_req_name = EXTERNAL_RUNTIME_REQ_MAP.get(req.name, req.name)
 
-    if not runtime_in_upstream_requires(
-        req, data
-    ) and not runtime_in_upstream_sdist_requires(req, data):
+    if (
+        runtime_req_name != upstream_distribution  # Allow `types-foo` to require `foo`
```
Avasam, line 354: if you want me to hide this check behind a function for consistency, I'd call it `runtime_is_upstream`.

srittau, line 354: I think this is the bug that I alluded to when I wrote the refactoring PR. But wouldn't it be clearer to just check `req.name == upstream_distribution` at the beginning of the function and exit early? This would also skip fetching the data from PyPI.

Avasam: If we don't mind not validating that `upstream_distribution` does exist on pypi

Avasam: > I think this is the bug that I alluded to when I wrote the refactoring PR. But wouldn't it be clearer to just check `req.name == upstream_distribution`

I'm not sure which side the the `EXTERNAL_RUNTIME_REQ_MAP` makes sense here. But I was reusing what's in `runtime_in_upstream_sdist_requires` and `runtime_in_upstream_requires`

PR #160, opened 2025-02-14
setuptools 69 includes type information by default

### LINKED ISSUE #112: setuptools v69 includes type information by default

Avasam:
Closes #112


tests/test_unit.py
```diff
@@ -117,38 +117,15 @@ def test_collect_package_data() -> None:
     stubs = Path("data") / "test_typeshed" / "stubs"
     pkg_data = collect_package_data(stubs / "singlefilepkg")
     assert pkg_data.top_level_packages == ["singlefilepkg-stubs"]
-    assert pkg_data.package_data == (
-        {"singlefilepkg-stubs": ["__init__.pyi", "METADATA.toml", "py.typed"]}
-    )
+    assert pkg_data.package_data == ({"singlefilepkg-stubs": ["METADATA.toml"]})
```
Avasam, line 120: Tests were checking for strings in `package_data`, but I don't think there's any test to ensure that the right files are *actually* collected.

stub_uploader/build_wheel.py
```diff
@@ -307,18 +301,13 @@ def collect_package_data(base_path: Path) -> PackageData:
...
         elif entry.is_dir():
             pkg_name = entry.name
-            stub_files = find_stub_files(str(entry))
+            validate_only_stub_files(str(entry))
         else:
             raise ValueError(f"Only stub files are allowed, not {entry.name!r}")
-        package_data[pkg_name + SUFFIX] = [*stub_files, META]
-    pkg_data = PackageData(base_path, package_data)
-    for package in pkg_data.top_level_non_namespace_packages:
-        pkg_data.add_file(package, "py.typed")
```
srittau, line 320: As far as I can see, this is the only call of `PackageData.add_file()`, so we can remove that method.

stub_uploader/build_wheel.py
```diff
@@ -216,30 +216,24 @@ def is_namespace_package(path: Path) -> bool:
...
+    for _, _, files in os.walk(top):
         for file in files:
             if file.endswith(".pyi"):
                 name, _ = os.path.splitext(file)
-                assert (
-                    name.isidentifier()
-                ), "All file names must be valid Python modules"
-                result.append(os.path.relpath(os.path.join(root, file), top))
-            elif file == "py.typed":
-                result.append(os.path.relpath(os.path.join(root, file), top))
```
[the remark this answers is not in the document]
Avasam: Pretty sure this only checks typeshed's content.

`validate_only_stub_files` is only called by `collect_package_data`. Which outside of tests, is called in only one place and before `create_py_typed`:
image

tests/test_unit.py
Avasam: This would be especially important to test if we swap build backends (https://github.com/typeshed-internal/stub_uploader/issues/166)

srittau (approved):
Just confirmed that there are only expected changes between the builds. Thanks!

srittau:
https://pypi.org/project/types-ExifRead/

srittau:
Reverted in #167, since this can't deal with namespace packages.

PR #161, opened 2025-02-25
Refactor verify_external_req for readability

srittau:
While trying to review #159, I found the `verify_external_req()` function very hard to understand and to follow its logic. It has multiple early exits, and understanding what the individual checks do (and where they end) is not trivial. In fact, I think a bug or two hides in the function.

Therefore, this PR extracts the individual checks, and reorders the logic to prevent early exist to make it easier to follow.

@Avasam: Do you want to review this? After this is merged, you'd have to adapt #159, but I promise to review it promptly.

### DIFF
--- a/stub_uploader/metadata.py
+++ b/stub_uploader/metadata.py
@@ -197,7 +197,7 @@ def verify_typeshed_req(req: Requirement) -> None:
     if not req.name.startswith(TYPES_PREFIX):
         raise InvalidRequires(f"Expected dependency {req} to start with {TYPES_PREFIX}")
 
-    if not canonical_name(req.name) in uploaded_packages.read():
+    if canonical_name(req.name) not in uploaded_packages.read():
         raise InvalidRequires(
             f"Expected dependency {req} to be uploaded from stub_uploader"
         )
@@ -234,7 +234,8 @@ EXTERNAL_REQ_ALLOWLIST = {
     "urllib3",
 }
 
-# Runtime requirements corresponding to the external requirements list above.
+# Map of external stub packages to their runtime equivalent. typeshed stubs
+# will depend on the stubs package,
 EXTERNAL_RUNTIME_REQ_MAP = {
     "django-stubs": "django",
     "djangorestframework-stubs": "djangorestframework",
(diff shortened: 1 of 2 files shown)


srittau (commented):
Finish comment

Avasam:
I'll be really occupied for the next 2 days. So it depends if I find the energy. But can review by the end of the week if no one beats me to it!

stub_uploader/metadata.py
```diff
@@ -283,23 +284,37 @@ def verify_external_req(
     upstream_distribution: Optional[str],
     _unsafe_ignore_allowlist: bool = False,  # used for tests
```
Avasam, line 285: You didn't touch this code, but I think the same as my comment below could apply:
```suggestion
    *,
    _unsafe_ignore_allowlist: bool = False,  # used for tests
```

stub_uploader/metadata.py
```diff
@@ -283,23 +284,37 @@ def verify_external_req(
...
         )
 
-    if upstream_distribution is None:
-        raise InvalidRequires(
-            f"There is no upstream distribution on PyPI, so cannot verify {req}"
-        )
 
+def verify_external_req_in_allowlist(
+    req: Requirement,
+    _unsafe_ignore_allowlist: bool = False,  # used for tests
```
Avasam, line 316: Let's take a page out of flake8-boolean-trap's book to help ensure this isn't used accidentally and is easily recognizable from the call site:
```suggestion
    *,
    _unsafe_ignore_allowlist: bool = False,  # used for tests
```

stub_uploader/metadata.py
```diff
@@ -234,15 +234,16 @@ def verify_typeshed_req(req: Requirement) -> None:
     "urllib3",
 }
 
-# Runtime requirements corresponding to the external requirements list above.
+# Map of external stub packages to their runtime equivalent. We check that
+# the stubs actually depend on their runtime package.
```
Avasam, line 238: Nitpick, but this comment feels like it split in an odd place when it wouldn't have to be
```suggestion
# Map of external stub packages to their runtime equivalent.
# We check that the stubs actually depend on their runtime package.
```

Avasam (commented):
This is a nice split of a complicated function !
I also appreciate the name change to `validate_pypi_response` to make it less ambiguous.

I have 1 suggested improvement and 1 nitpick. Neither are blockers. Otherwise LGTM.

PR #164, opened 2025-03-29
Remove the license trove classifier

srittau:
PEP 639 deprecates the classifier and setuptools started to issue a warning if it's used:
https://peps.python.org/pep-0639/#deprecate-license-classifiers

Closes: python/typeshed#13738

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -76,7 +76,6 @@ setup(name=name,
       license="Apache-2.0",
       python_requires="{requires_python}",
       classifiers=[
-          "License :: OSI Approved :: Apache Software License",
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):

cdce8p:
I'm currently working on moving the project config from `setup.py` to `pyproject.toml`. With the latest setuptools releases, this would allow for PEP 639 (license expressions).

It's currently blocked by a small bug which prevents the use of `package-data` for stubs packages but that should be resolved soon. https://github.com/abravalheri/validate-pyproject/pull/248

cdce8p:
Update: Just created https://github.com/typeshed-internal/stub_uploader/pull/165 for everything except `package-data`.

PR #167, opened 2025-04-02
Revert "setuptools 69 includes type information by default (#160)"

srittau:
This reverts commit 1fe4ba9734daeacc998861576ac824f976530dc3.


srittau:
See python/typeshed#13770 for background.

PR #169, opened 2025-04-02
Use a prefix for the tmp build directory

srittau:


srittau:
This allows me to clearly distinguish those directories in my overfull `/tmp` directory.

PR #170, opened 2025-04-03
Integration tests: Check for files existing

srittau:
In particular, this checks whether namespace packages
include *.pyi files.

* Return `Path` not `str` from `build_wheel.main()`.
* Delete temporary directories while running integration tests.


srittau:
Bumpity bump.

ilevkivskyi (approved):

PR #173, opened 2025-05-04
Add asgiref to allowlist

huynguyengl99:
Prerequisite for [python/typeshed#13939](https://github.com/python/typeshed/pull/13939)

* [Django channels stubs python/typeshed#13939](https://github.com/python/typeshed/pull/13939)

The package is in the [PyPI top 1000 list](https://hugovk.github.io/top-pypi-packages/) (ranked 195th).  It is fully type-hinted and under the Django umbrella, so I think it’s safe to add to the EXTERNAL_REQ_ALLOWLIST.


srittau (changes requested):
Sounds reasonable, but please keep the list sorted.

huynguyengl99:
Oh, sorry about that! I didn’t realize the order — I updated it.

srittau (approved):

PR #176, opened 2025-07-16
Remove support for pytype

srittau:
Also improve the generated package descriptions:

* List type checkers using a markdown list.
* Link the type checkers in that list.
* Instead remove the mention of specific type checkers from the
  introduction. (This removes mentions of pyre and PyCharm,
  but I assume that at some point we will add pyrefly to the supported
  type checkers anyway.)
* Use "type checkers" instead of "type-checking tools".

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -2,13 +2,8 @@
 Basic script to generate a wheel for a third-party distribution in typeshed.
 
 This generates a PEP 561 types stub package using METADATA.toml file for a given
-distribution in typeshed stubs. Such package
-can be used by type-checking tools like
-[mypy](https://github.com/python/mypy/),
-[pyright](https://github.com/microsoft/pyright),
-[pytype](https://github.com/google/pytype/),
-PyCharm, etc. to check code that uses
-the corresponding runtime Python package.
+distribution in typeshed stubs. Such package can be used by type checkers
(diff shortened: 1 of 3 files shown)


srittau:
Cf. python/typeshed#11958

stub_uploader/build_wheel.py
```diff
@@ -102,14 +97,9 @@
...
 
-This is a [PEP 561](https://peps.python.org/pep-0561/)
-type stub package for the {formatted_distribution} package.
-It can be used by type-checking tools like
-[mypy](https://github.com/python/mypy/),
-[pyright](https://github.com/microsoft/pyright),
-[pytype](https://github.com/google/pytype/),
-[Pyre](https://pyre-check.org/),
-PyCharm, etc. to check code that uses `{distribution}`. This version of
+This is a [PEP 561](https://peps.python.org/pep-0561/) type stub package for
```
JelleZijlstra, line 100: What do you think of linking to the spec (https://typing.python.org/en/latest/spec/distributing.html#type-information-in-libraries) instead of PEP 561? PEPs are not ongoing documentation after all.

srittau: There's also a link to "partial" that needs to be replaced. But let me do that in a followup PR.

srittau:
Pingety pong. I'd like to finish this in typeshed.

JelleZijlstra (approved):

PR #180, opened 2025-08-09
Generate links for typeshed PR numbers in changelog entries

### LINKED ISSUE #178: Link PR numbers in changelog entries?
It would be nice if changelog entries somehow linked back to the corresponding typeshed PRs. Currently, if a user wants more information about a changelog entry, they need to manually copy the PR number and search for it on the typeshed repo.

One idea would be to search for the pattern `(#NNNNN)` in new changelog entries and replace it with something like `(python/typeshed#NNNNN)` or `([#NNNNN](https://github.com/python/typeshed/pull/NNNNN))`.

brianschubert:
Closes #178

### DIFF
--- a/stub_uploader/update_changelog.py
+++ b/stub_uploader/update_changelog.py
@@ -62,7 +62,8 @@ def update_changelog(
     if not log:
         print(f"{distribution}: Changelog unchanged")
         return
-    new_entry = process_git_log(log, version)
+    today = datetime.date.today()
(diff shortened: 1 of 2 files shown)


srittau:
This should exclude items where there is text in front of the hash. E.g. `foo/bar#123` should not be linked. This might be as easy as adding `(^|\W)` to the front of the regex.

srittau (approved):

PR #196, opened 2026-02-04
Prettify changelogs

donbarbos:
* Remove scope prefix for commit titles (It doesn't seem to make sense for package changelog)
* Add indentation to commit description
* Add ' \\' for correct line breaks
* Add link to PyPI release in header (This is debatable, and I’m not sure how useful this would be, as the raw Markdown may not be very readable)

### `Authlib.md` changelog example:

| After | Before |
|----------|--------|
| scrn21 | scrn22 |

| After | Before |
|----------|--------|
| scrn31 | scrn32 |

### DIFF
--- a/stub_uploader/update_changelog.py
+++ b/stub_uploader/update_changelog.py
@@ -13,6 +13,7 @@ from pathlib import Path
 
 from stub_uploader.const import CHANGELOG_PATH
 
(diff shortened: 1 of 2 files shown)


neeehvns-tech (approved):

srittau (approved):
Sorry for the long delay. Looks great!

PR #200, opened 2026-04-05
use dashes instead of underscores

lev-blit:
the second part of https://github.com/python/typeshed/issues/15550


lev-blit:
@srittau once this merges I'll open the second PR for typeshed :)

brianschubert:
There's also a fixture that needs to be updated here: https://github.com/typeshed-internal/stub_uploader/blob/24bf940cd7ad70fc3db7ce625af9dd2c956bb54a/tests/test_unit.py#L314-L317

srittau (approved):
Nice, thanks!

PR #201, opened 2026-04-28
Remove METADATA field name fallbacks

srittau:


srittau:
@donbarbos Maybe you want to have a look?

donbarbos (commented):
There is another fallback with TODO comment: https://github.com/typeshed-internal/stub_uploader/blob/c00229299742926f897090002c09434e51636caf/stub_uploader/ts_data.py#L44-L47

PR #202, opened 2026-04-28
Support structured obsolete-since field

srittau:
Builds on #201.


srittau:
@donbarbos Since you worked on this recently, maybe you're interested in reviewing this and its companion python/typeshed#15682?

PR #210, opened 2026-07-26
Generate '[project.optional-dependencies]' table for pyproject.toml

donbarbos:


srittau (approved):
I couldn't decide which key to use here, which is why I originally left the optional dependencies out here. But this is a good solution!

PR #213, opened 2026-07-27
Unpin requests

srittau:
It was originally pinned in #95 in 2023, because of CI failures with the new (at the time) requests 2.30.0. This was due to some interaction of requests and urllib3 (see psf/requests#6439). This has been fixed for a long time.

Cc @AlexWaygood

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # scripts
 packaging
-requests < 2.30.0
+requests
 
 # build and upload
 build


AlexWaygood (approved):
I feel like a better long-term model here would be to pin/lock these dependencies and have automated updates by renovate or dependabot, since this is not a library. That would prevent unexpected CI failures due to new releases of libraries we depend on. But that's obviously out of scope for this PR!

srittau:
I agree. I'd also like to switch to pyproject.toml instead of requirements. Well, when there's time ... (In the meantime, I'll open an issue.)

PR #216, opened 2026-07-29
Update URLs for mypy and pyright

srittau:
Use the marketing URLs, not the GitHub repository

Depends on #214


srittau:
This was closed for unknown reasons and can't be reopened? GitHub is weird sometimes.

PR #217, opened 2026-07-29
Update URLs for mypy and pyright

srittau:
Link to marketing sites instead of GitHub.


JelleZijlstra:
Personally I find the GitHub repo links more useful; won't block this change though

srittau:
I find the project sites more useful for the people who read those descriptions: users wanting to find our more about the type checkers in question; GitHub isn't that useful if you just want to use a type checkers.

But if we decide to use GitHub links we need to change ty's link as well to stay consistent.

AlexWaygood (approved):
I also find GitHub links more useful as a maintainer, but I think most users probably want to see the rendered docs more than the implementation, and the marketing links are probably better for that

PR #218, opened 2026-08-01
Apply new rules to old changelogs

donbarbos:
Updated the entries after 15 Oct 2023 to account for the 02:00 upload time.

Since my script is based on heuristics (including the assumption that commits fall within the expected time window), I had to verify the results manually. The fact that the numbers of git additions and deletions match gives me confidence that the update was successful.

<details>
  <summary>I used next script:</summary>

```python
import re
import subprocess
import os
import datetime
from dataclasses import dataclass
from pathlib import Path

from stub_uploader.const import CHANGELOG_PATH, THIRD_PARTY_NAMESPACE
from stub_uploader.update_changelog import process_git_log

# First auto commit planned at 02:00:00 was merged on Oct 16 2023
# See typeshed-internal/stub_uploader/pull/106
CUTOFF_DATE = datetime.date(2023, 10, 16)
TYPESHED_DIR = "../typeshed"
OLD_HEADER_RE = re.compile(
    r"^## (?P<version>\S+) \((?P<date>\d{4}-\d{2}-\d{2})\)$",
)
NEW_HEADER_RE = re.compile(
    r"^## \[(?P<version>[^\]]+)\]\(https://pypi\.org/project/types-[^)]+\) "
    r"\((?P<date>\d{4}-\d{2}-\d{2})\)$",
)
HEADER_RE = re.compile(r"^## .*?$", re.MULTILINE)


@dataclass
class ChangelogEntry:
    version: str
    date: datetime.date
    header: str
    body: str
    original: str
    is_new: bool


def parse_entries(content: str) -> list[ChangelogEntry]:
    matches = list(HEADER_RE.finditer(content))
    entries: list[ChangelogEntry] = []

    for i, match in enumerate(matches):
        start = match.start()
        end = matches[i + 1].start() if i + 1 < len(matches) else len(content)

        original = content[start:end]
        header = match.group(0)

        m = NEW_HEADER_RE.match(header)
        is_new = m is not None

        if m is None:
            m = OLD_HEADER_RE.match(header)

        if m is None:
            raise RuntimeError(f"Cannot parse header:\n{header}")

        entries.append(
            ChangelogEntry(
                version=m["version"],
                date=datetime.date.fromisoformat(m["date"]),
                header=header,
                body=original[len(header) :],
                original=original,
                is_new=is_new,
            )
        )

    return entries


def git_log_between(
    *,
    typeshed_dir: str,
    distribution: str,
    before: datetime.date,
    after: datetime.date,
) -> str:
    dist_path = Path(THIRD_PARTY_NAMESPACE, distribution)
    result = subprocess.run(
        [
            "git",
            "log",
            "--before",
            f"{before}T05:00:00+00:00",  # 5am is approximately the auto commit time.
            "--after",
            f"{after}T05:00:00+00:00",
            "--",
            dist_path,
        ],
        cwd=typeshed_dir,
        stdout=subprocess.PIPE,
        check=True,
    )

    return result.stdout.decode("utf-8")


def regenerate_changelog(content: str, distribution: str) -> str:
    entries = parse_entries(content)

    result: list[str] = []

    for i, entry in enumerate(entries):
        if entry.is_new:
            result.append(entry.original)
            continue

        if entry.date <= CUTOFF_DATE:
            result.extend(e.original for e in entries[i:])
            break

        next_entry_date = entries[i + 1].date if i + 1 < len(entries) else CUTOFF_DATE
        if next_entry_date < CUTOFF_DATE:
            next_entry_date = CUTOFF_DATE

        log = git_log_between(
            typeshed_dir=TYPESHED_DIR,
            distribution=distribution,
            before=entry.date,  # For example: 20 Dec
            after=next_entry_date,  # For example: 16 Oct
        )

        if not log.strip():
            print(
                f"WARNING: empty git log for {distribution} {entry.version} ({entry.date})"
            )
            result.append(entry.original)
            continue

        result.append(
            process_git_log(
                log=log,
                distribution=distribution,
                version=entry.version,
                today=entry.date,
            )
        )

    return "".join(result)


def main() -> None:
    for filename in os.listdir(CHANGELOG_PATH):
        if not filename.endswith(".md"):
            continue
        changelog = CHANGELOG_PATH / filename

        with open(changelog, "r", encoding="utf-8") as f:
            content = f.read()

        distribution = changelog.stem
        updated_content = regenerate_changelog(content, distribution)

        with open(changelog, "w", encoding="utf-8") as f:
            f.write(updated_content)


if __name__ == "__main__":
    main()
```
</details>


data/changelogs/docker.md
```diff
@@ -14,320 +14,320 @@
...
-            chunk['stream']
-        )
-        if match:
-            image_id = match.group(2)
-    last_event = chunk
-if image_id:
-    return (self.get(image_id), result_stream)
-raise BuildError(last_event or 'Unknown', result_stream)
-```
+    ```python \
```
ilevkivskyi: Are these backslashes necessary? Also some of the backslashes elsewhere look redundant, but maybe I am missing something.

ilevkivskyi (commented):
In general, I am skeptical of the changes of this size, as it is very easy to miss something. However, I am not going to block this if some other core team member wants to review, and take this under their responsibility.

data/changelogs/docker.md
donbarbos: Indeed, the backslash isn't actually necessary.
I originally added this rule in PR #196 to preserve line breaks when there isn't a blank line separating paragraphs in the commit body. This is useful because commit messages often intentionally use line breaks for descriptions such as:
```
info: someurl
diff: anotherurl
key: value
```
Without the backslash, markdown would render those lines as a single line. It seems worth being a bit more careful about when inserting the backslash in `process_git_log`.

donbarbos: Fixed #219

PR #219, opened 2026-08-02
Fix backslash rule for process_git_log

donbarbos:
See: https://github.com/typeshed-internal/stub_uploader/pull/218#discussion_r3698773223


ilevkivskyi (approved):
This will still trigger on two random sentences with a colon in each, but I guess this is good enough (and at least better than before).

PR #226, opened 2026-09-01
Run linters/formatters via pre-commit

AlexWaygood:
This will make it easier to add more security and correctness linters to CI in followup PRs. This is a similar pattern to what we already use in typeshed and typing_extensions

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -13,7 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # 6.0.3
-      - uses: j178/prek-action@5337cb91e0fa35a7ff31b9ca345126d8bbbcdf16 # v2.0.6
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # 6.2.0
+        with:
+          python-version: '3.13'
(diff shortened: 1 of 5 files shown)


.github/workflows/check.yml
```diff
@@ -8,18 +8,12 @@ env:
...
     steps:
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # 6.2.0
-        with:
-          python-version: '3.13'
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # 6.0.3
-      - run: pip install -r requirements.txt
-      - run: |
-          black --version
-          black --check .
+      - uses: j178/prek-action@5337cb91e0fa35a7ff31b9ca345126d8bbbcdf16 # v2.0.6
```
srittau: Can't we use the "real" pre-commit here? This sounds safer to me than a replacement by an anonymous GitHub user.

AlexWaygood: of course. I default to prek these days since it's so much faster, but very happy to switch back to pre-commit here

srittau: If you trust prek, I trust your judgement, but pre-commit sounds like the safer alternative to me, especially in this fairly critical project.

AlexWaygood: yeah, I agree with your assessment here

srittau (approved):

PR #227, opened 2026-09-01
Add zizmor hook and fix workflow security findings

AlexWaygood:
Add [zizmor](https://docs.zizmor.sh/) as a SHA-pinned pre-commit hook and address its findings in our workflows. Zizmor statically checks GitHub Actions configurations for security issues including script injection, excessive token permissions, credential leakage, dangerous workflow triggers, and unpinned or known-vulnerable actions.

The [`template-injection`](https://docs.zizmor.sh/audits/#template-injection) fixes affect two manually dispatched publishing workflows:

- In `force_update.yml`, move `github.event.inputs.distribution` into the step's `DISTRIBUTION` environment variable and pass it to `stub_uploader.upload_some` as `"$DISTRIBUTION"`.
- In `test_api_token.yml`, move `github.event.inputs.increment` into the step's `PACKAGE_VERSION_INCREMENT` environment variable, which the package build already reads from its environment.

Previously, GitHub substituted these inputs directly into the `run` scripts before the shell parsed them. A supplied value containing shell syntax, such as a command substitution, could therefore execute commands in a step with access to the PyPI token; surrounding a GitHub expression with double quotes does not prevent that. Passing inputs through `env` keeps their contents out of the shell source. The quoted `"$DISTRIBUTION"` expansion passes the value as a single argument without interpreting its contents as shell commands, while the version increment reaches the build directly through its environment.

The remaining changes restrict GitHub token permissions to `contents: read`, granting `contents: write` only to jobs that push upload records, and disable checkout credential persistence except where those pushes require it. The API-token test has a documented exception to `use-trusted-publishing`, since switching that workflow to trusted publishing would defeat its purpose.

Action version comments now include the tags' `v` prefix, for example `# v6.0.3`, so zizmor can match them to the existing SHA pins. If we enable Dependabot or Renovate for this repository in the future, either will still update both the pinned action SHA and the adjacent version comment in this format. Both [Dependabot](https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#github-actions) and [Renovate](https://docs.renovatebot.com/modules/manager/github-actions/#digest-pinning-and-updating) support it.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -4,6 +4,9 @@ on:
   push:
   pull_request:
 
+permissions:
+  contents: read
+
 env:
   FORCE_COLOR: 1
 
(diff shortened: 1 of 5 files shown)


JelleZijlstra (approved):

AlexWaygood:
(I can't merge this myself as I'm not a maintainer on this repo)

PR #228, opened 2026-09-02
Add httpx2 to the external dependency allowlist

donbarbos:
Source: https://github.com/python/typeshed/pull/16326#discussion_r3893918387


srittau:
We should also be able to remove `httpx`, as it's not used anymore, I think.

srittau (approved):
Thanks!

PR #229, opened 2026-09-05
Document optional dependencies for stub packages

donbarbos:
* Example of generated list:
````
```console
pip install 'types-fake[all]' # installs all optional dependencies
pip install 'types-fake[numpy]' # installs numpy
pip install 'types-fake[requests]' # installs types-requests>=0.1
pip install 'types-fake[pandas]' # installs pandas-stubs
```
````
* Sort description templates by the order in which they are used

### DIFF
--- a/stub_uploader/build_wheel.py
+++ b/stub_uploader/build_wheel.py
@@ -44,11 +44,6 @@ CHANGELOG = "CHANGELOG.md"
 
 SUFFIX = "-stubs"
 
-PARTIAL_STUBS_DESCRIPTION = """
-This stub package is marked as [partial](https://typing.python.org/en/latest/spec/distributing.html#partial-stub-packages).
-If you find that annotations are missing, feel free to contribute and help complete them.
-""".lstrip()
-
 PYPROJECT_TEMPLATE = dedent("""
 [build-system]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: stub_uploader/build_wheel.py [hunk 1]
```diff
@@ -120,6 +120,18 @@
...
 """.strip()
 
+EXTRAS_SECTION_TEMPLATE = """
+## Optional dependencies
+
+Some stubs depend on third-party libraries that are not required by every user.
+The following extras are available:
+
+```console
+pip install '{stub_distribution}[all]' # installs all optional dependencies
```
srittau:
```suggestion
- pip install '{stub_distribution}[all]' # installs all optional dependencies
```

stub_uploader/build_wheel.py
```diff
@@ -433,6 +445,18 @@ def generate_long_description(
             ),
         )
     )
+    if metadata.optional_dependencies:
+        parts.append(
+            EXTRAS_SECTION_TEMPLATE.format(
+                stub_distribution=metadata.stub_distribution,
+                extras_install_commands="\n".join(
+                    f"pip install '{metadata.stub_distribution}"
```
srittau:
```suggestion
                    f"- pip install '{metadata.stub_distribution}"
```

srittau (changes requested):
To ensure correct rendering as a list

stub_uploader/build_wheel.py
donbarbos: To be honest, I'm not quite sure what the purpose of this formatting is. It will be rendered together with the `- ` at the beginning of the line. Or are you trying to indicate that this is a console command? In that case, perhaps we could add `$ ` before the command instead, e.g.
```
$ pip install ...
$ pip install ...
$ pip install ...
```

srittau: My bad, I overlooked the `console` marker before that. I'm fine with either the original rendering or the `$` marker.

srittau (approved):

PR #231, opened 2026-09-06
Lock direct and indirect dependencies using uv

AlexWaygood:
Security is pretty important for this repo, but most of our direct dependencies, and all of our indirect dependencies, are currently entirely unpinned, which isn't great from a security perspective. This PR adds a `uv.lock` file so that all direct and indirect dependencies are pinned both in CI and when developing the project locally.

In general, I'm not a huge fan of mandating that a particular package-management tool has to be used when developing an OSS project locally. I don't particularly appreciate being told that I have to download some tool I don't generally use in order to contribute to an OSS project, and don't _love_ putting other folks in that position.

However, this repo doesn't really get many external contributors (it's almost exclusively maintained by the typeshed/mypy teams); pinning dependencies is extremely appropriate for this repo (since it's more of an "application" codebase than a "library" codebase); and security is an even higher concern than in most repos. So this PR proposes adding a uv lockfile.

We don't have to use uv, of course. We could use another package-management tool like pdm, hatch, etc., that allows indirect dependencies to be pinned. I'm biased towards uv since I work on the Astral team and know well how it works, but I don't mind if folks prefer to use something else for this repo!


.github/workflows/check.yml
```diff
@@ -27,29 +27,31 @@ jobs:
...
     steps:
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
-        with:
-          python-version: '3.13'
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
           persist-credentials: false
-      - run: pip install -r requirements.txt
-      - run: mypy --strict -p stub_uploader -p tests
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
```
srittau, line 33: Considering that astral-sh/setup-uv uses immutable releases, we should be able to just use the version number:


```suggestion
      - uses: astral-sh/setup-uv@v10.0.1
```

.github/workflows/check.yml
```diff
@@ -27,29 +27,31 @@ jobs:
...
-        with:
-          # Keep in sync with typeshed's daily.yml and tests.yml workflows.
-          python-version: '3.13'
       - name: Checkout main
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
           path: main
           persist-credentials: false
+      - name: Setup uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
```
srittau, line 49:
```suggestion
        uses: astral-sh/setup-uv@v10.0.1
```

.github/workflows/force_update.yml
```diff
@@ -18,34 +18,36 @@ jobs:
...
-        with:
-          python-version: '3.13'
       - name: Checkout main
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
           path: main
           # Needed to push updates to the uploaded packages list.
           persist-credentials: true
+      - name: Setup uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
```
srittau, line 28:
```suggestion
        uses: astral-sh/setup-uv@v10.0.1
```

.github/workflows/test_api_token.yml
```diff
@@ -23,14 +23,14 @@ jobs:
     - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
       with:
         persist-credentials: false
-    - name: Set up Python
-      uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
+    - name: Setup uv
+      uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
```
srittau, line 27:
```suggestion
      uses: astral-sh/setup-uv@v10.0.1
```

.github/workflows/update_stubs.yml
```diff
@@ -19,16 +19,19 @@ jobs:
...
-        with:
-          python-version: '3.13'
       - name: Checkout main
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
           path: main
           # Needed to push updated commit records and changelogs.
           persist-credentials: true
+      - name: Setup uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
```
srittau, line 29:
```suggestion
        uses: astral-sh/setup-uv@v10.0.1
```

srittau (approved):
I'm in favor of this, but would like the feedback of @ilevkivskyi and @JelleZijlstra, as the other committers of this repository.

.github/workflows/check.yml
AlexWaygood: While that's true, I think we'd have to add a zizmor ignore if we just pinned to a tag here, which itself would need to be justified in a comment. And pinning to a hash means we can see just by glancing at the workflow that we're pinning the action in a secure way, rather than having to consult `setup-uv` and check whether it uses immutable releases or not. So personally I would prefer the explicit hash pin here.

JelleZijlstra (approved):

ilevkivskyi:
Not a huge fan, but also not against this. Hopefully, pip will support this natively soon.

